PR #49, opened 2022-01-29
Add ipv6_neighbor_discovery

rifelpet:
This seemed to work well during minimal testing, it could use a thorough review though.

I also couldn't get the doc generation to work, so you may need to add those.


GNewbury1:
So sorry for the delay, been having a very busy few months.

The change looks good. I can handle the doc generation, but the resource should have a test (you have one for the go client, but not the resource itself).

PR #56, opened 2022-07-01
/system/identity set/get

### LINKED ISSUE #55: Ability to manage system identity
Ability to manage `/system/identity`

michael-robbins:
Because we cannot 'delete' a System Identity we just reset it back to the default `MikroTik`. If that's not the best idea, we can instead just make the delete a 'no op' and skip it.

Closes #55


GNewbury1:
Sorry for the delay - been having a busy few months! This LGTM, I'll set the CI tests going

PR #58, opened 2022-08-30
GRE interface support

vaerh:
Hi, can you review and add GRE support?
And a few changes to remove the output of unnecessary information.


GNewbury1:
Hey @vaerh, I'm happy to review more changes if you want to re-open this pull request

vaerh:
Hello. I went with my changes further than the current architecture. Is there any desire to merge our repositories?
I've made it easy to add resources (now you only need to correctly and slowly describe the schema) without the need to create data storage structures.
If possible, I suggest looking at the code in my repository. And, if it does not cause rejection, think about how we should proceed.

Issue #73, opened 2022-10-23
Bridge interface: pvid out of range


JanNemcik:
**Describe the bug**
Not an expert for RouterOS but when I want to create a bridge interface without setting the pvid parameter, I'm getting following error: 
```bash
When applying changes to routeros_interface_bridge.bridge, provider "provider[\"registry.terraform.io/gnewbury1/routeros\"]" produced an unexpected new value: Root resource was present, but now absent.
```
After looking into Terraform debug logs I noticed following line:
```bash
{"@level":"warn","@message":"unexpected data","@timestamp":"2022-10-23T14:04:28.141949+02:00","registry.terraform.io/gnewbury1/routeros:stdout":"value of pvid out of range (1..4094)"}
```
Based on the error and according to the [docs[(https://help.mikrotik.com/docs/display/ROS/Bridging+and+Switching) I'm assuming that default value should be 1 and not 0 as in the current state:
```go
"pvid": {
    Type:     schema.TypeInt,
    Optional: true,
    Default:  0,
},
```

**To Reproduce**
Try to apply example configuration of _routeros_interface_bridge_

**Expected behavior**
Successful creation of bridge interface

**Stack Trace**
```bash
routeros_interface_bridge.bridge: Modifying... [id=*<id>]
....
{"@level":"trace","@message":"received data","@module":"provider.stdio","@timestamp":"2022-10-23T14:04:28.141913+02:00","channel":"STDOUT","len":36}
{"@level":"warn","@message":"unexpected data","@timestamp":"2022-10-23T14:04:28.141949+02:00","registry.terraform.io/gnewbury1/routeros:stdout":"value of pvid out of range (1..4094)"}
....
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to routeros_interface_bridge.bridge, provider "provider[\"registry.terraform.io/gnewbury1/routeros\"]" produced an unexpected new value: Root resource was present, but now absent.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

**Additional context**
Versions:
Terraform: 1.3.3
Provider: 0.3.4
RouterOS: 7.5

PR #74, opened 2022-10-23
fix: pvid defaults to 1

JanNemcik:
According to [offcial docs](https://help.mikrotik.com/docs/display/ROS/Bridging+and+Switching) pvid should be within range of 1 to 4094 defaulting to 1
Fixes: #73


GNewbury1:
Thanks for this pull request! Once the pipelines have run, I'll create a new patch release

GNewbury1:
Ah, the automated tests need to be updated as well. I'll be happy to do that in the morning, then release as 0.3.5

PR #94, opened 2023-02-12
Feature/update

GNewbury1:
Merge in changes from vaerh fork


vaerh:
Hello! Need my help to merge branches?

GNewbury1:
> Hello! Need my help to merge branches?

Possibly - I'm just going through a lot of the functionality that you replace (read: improved) and understand it. I think you're a much better Go developer than I am :P

vaerh:
Well, then I'll wait until you look at the code and then we need to think about how to merge it, since some points break compatibility in terraform scenarios.
All schema processing logic is contained in the following files: mikrotik*.go
I proceeded from the idea that it is possible to extract data types from the TF schema and transparently receive / give them to Mikrotik. This scenario covers most of the implemented resources so far, but in some cases it may be necessary to describe them in the old way.

GNewbury1:
The tests have finally passed! :D I'm going to update the README and run tfplugindocs before merging. As it's now gone midnight, I'll leave that until tomorrow. Then we can release a new "unified" version of the provider. Would be happy to have you onboard as a maintainer, too, if you're up for it

vaerh:
Good news, I will try to help to the best of my knowledge and ability!

PR #100, opened 2023-02-15
fix: Interface child items had incorrect reference

GNewbury1:


vaerh:
I just wanted to discuss the different resource naming in our versions. I have shortened some of the name. Maybe it makes sense for these resources to make a double naming?
https://github.com/GNewbury1/terraform-provider-routeros/blob/1ebca01793f8d423fbcc2fee33ae2de5dbc98a08/routeros/provider.go#L78-L85

GNewbury1:
> I just wanted to discuss the different resource naming in our versions. I have shortened some of the name. Maybe it makes sense for these resources to make a double naming?
> 
> https://github.com/GNewbury1/terraform-provider-routeros/blob/1ebca01793f8d423fbcc2fee33ae2de5dbc98a08/routeros/provider.go#L78-L85

Actually yeah, that's a good idea. Gimme a sec and I'll push the change for you to review

GNewbury1:
How does that look? Added some comments so we don't forget why we did this :P

vaerh (approved):
Exactly what is needed!

PR #173, opened 2023-04-12
fix: Remove default for VRRP interface group

jfroche:
Using RouterOS 7.7 we are facing errors related to the group used by multiple VRRP interface:

```
Error: from RouterOS device: input does not match any value of group-master
with routeros_interface_vrrp.interface_vrrp["4"],
on vrrp.tf line 3, in resource "routeros_interface_vrrp" "interface_vrrp":
  3: resource "routeros_interface_vrrp" "interface_vrrp" {
```

when declaring a VRRP interface:

```
resource "routeros_interface_vrrp" "interface_vrrp" {
  for_each  = { for interface in local.interfaces : interface.interface_id => interface }
  interface = "ether${each.value.interface_id}"
  name      = "vrrp${each.value.interface_id}"
  comment   = "${each.value.comment}_gw"
  priority  = "200"
  vrid      = "42"
}
```

This change remove the default value to avoid providing a group name on the interface to the API.

The request to the API was:

```
request body: /interface/vrrp/add =group-master= =comment=TEST_STAGING_gw =interval=1s =arp-timeout=auto =disabled=no =priority=200 =arp=enabled =vrid=42 =preemption-mode=yes =authentication=none =name=vrrp13 =version=3 =interface=ether13 =sync-connection-tracking=no =v3-protocol=ipv4
```

The request to the API is now:

```
request body: /interface/vrrp/add =comment=TEST_STAGING_gw =interval=1s =arp-timeout=auto =disabled=no =priority=200 =arp=enabled =vrid=42 =preemption-mode=yes =authentication=none =name=vrrp13 =version=3 =interface=ether13 =sync-connection-tracking=no =v3-protocol=ipv4
```

### DIFF
--- a/routeros/resource_interface_vrrp.go
+++ b/routeros/resource_interface_vrrp.go
@@ -56,7 +56,7 @@ func ResourceInterfaceVrrp() *schema.Resource {
 		"group_master": {
 			Type:        schema.TypeString,
 			Optional:    true,
-			Default:     "none",
+			Computed:    true,
 			Description: "Allows combining multiple VRRP interfaces to maintain the same VRRP status within the group.",
 			// Maybe this is a bug, but for the 'none' value, the Mikrotik ROS 7.5 returns an empty string.
 			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {


vaerh:
Hi, thanks for the fix. I think that this configuration may result in a non-empty plan. We need to check if Mikrotik returns something in this parameter or not.

jfroche:
Further plans contains no changes so it looks like the provider is already treating the empty string as undefined value.

Here the debug output during the plan after the vrrp interface creation:

```
2023-04-12T12:55:56.577+0200 [DEBUG] provider.terraform-provider-routeros_1.3.1: request body:  /interface/vrrp/print ?name=vrrp12: @caller=github.com/terraform-routeros/terraform-provider-routeros/routeros/log.go:16 @module=routeros tf_provider_addr=terraform-routeros/routeros tf_req_id=dc45742a-83fb-b3be-ea61-5b3990b70daa tf_rpc=Configure timestamp=2023-04-12T12:55:56.577+0200
2023-04-12T12:55:56.593+0200 [DEBUG] provider.terraform-provider-routeros_1.3.1: response body: !re @r27 [{`.id` `*2E`} {`name` `vrrp12`}
 {`mtu` `1500`} {`mac-address` `00:00:5E:00:01:2A`} {`arp` `enabled`} {`arp-timeout` `auto`} {`interface` `ether12`} {`group-master` ``}
{`vrid` `42`} {`priority` `200`} {`interval` `1s`} {`preemption-mode` `true`} {`authentication` `none`} {`password` ``} {`on-backup` ``}
{`on-master` ``} {`on-fail` ``} {`version` `3`} {`v3-protocol` `ipv4`} {`sync-connection-tracking` `false`} {`invalid` `true`} {`running`
 `false`} {`disabled` `false`} {`comment` `TELESERVICES_STAGING_gw`}]
 ```

File: routeros/resource_interface_vrrp.go [hunk 1]
```diff
@@ -56,7 +56,6 @@ func ResourceInterfaceVrrp() *schema.Resource {
 		"group_master": {
 			Type:        schema.TypeString,
 			Optional:    true,
-			Default:     "none",
```
vaerh, line 59: Please check the effect with the "Computed: true," parameter. I don't have a router at hand right now.

vaerh (changes requested):

routeros/resource_interface_vrrp.go
jfroche: Right the default value should come from the router API if nothing is provided by the user.
I have tested the change here and it works.

vaerh (approved):

PR #248, opened 2023-09-03
Document allowed url schemes

davidkern:
Small change to the provider docs to list out the allowable url schemes.

This caught me up for a moment since `http` isn't actually allowed. I documented only the `api` and `apis` schemes, leaving `https` undocumented.

`https`, while allowed, is confusing in its own way since without a port specified 443 would be assumed by most, not 8729.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,7 +24,7 @@ terraform {
 }
 
 provider "routeros" {
-  hosturl        = "https://router.local"        # env ROS_HOSTURL or MIKROTIK_HOST
+  hosturl        = "apis://router.local"         # env ROS_HOSTURL or MIKROTIK_HOST
   username       = "admin"                       # env ROS_USERNAME or MIKROTIK_USER
   password       = ""                            # env ROS_PASSWORD or MIKROTIK_PASSWORD
   ca_certificate = "/path/to/ca/certificate.pem" # env ROS_CA_CERTIFICATE or MIKROTIK_CA_CERTIFICATE
@@ -43,7 +43,9 @@ resource "routeros_interface_gre" "gre_hq" {
 
 ### Required
 
-- `hosturl` (String) URL of the ROS router. Include the scheme (http/https)
+- `hosturl` (String) URL of the ROS router including the scheme:
+  - `api` http access on port 8728
+  - `apis` https access on port 8729
 
 ### Optional
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,7 +24,7 @@ terraform {
 }
 
 provider "routeros" {
-  hosturl        = "apis://router.local"         # env ROS_HOSTURL or MIKROTIK_HOST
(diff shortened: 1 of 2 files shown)


vaerh:
Hey, David. 
Thanks for the corrections. Absolutely agree about http, as the REST API only works via https. 
I would like to point out that the REST API works via port 443.

docs/index.md
```diff
@@ -24,7 +24,7 @@ terraform {
 }
 
 provider "routeros" {
-  hosturl        = "https://router.local"        # env ROS_HOSTURL or MIKROTIK_HOST
+  hosturl        = "apis://router.local"         # env ROS_HOSTURL or MIKROTIK_HOST
```
vaerh, line 27: The documentation is auto-generated. Please make additions to https://github.com/terraform-routeros/terraform-provider-routeros/blob/main/examples/provider/provider.tf instead of this file.
And please do not remove the existing parameter, but add your own under the comment # so that the alternative is visible.  A lot of people manage routers via https.

docs/index.md
```diff
@@ -43,7 +43,9 @@ resource "routeros_interface_gre" "gre_hq" {
 
 ### Required
 
-- `hosturl` (String) URL of the ROS router. Include the scheme (http/https)
```
vaerh, line 46: That's a bit of a misnomer:
api is the old API without TLS/SSL
apis is old API with TLS/SSL
https is new REST API with TLS/SSL. All three schemes are supported with their own separate ports. So please do not exclude https. 
Correct it in the file https://github.com/terraform-routeros/terraform-provider-routeros/blob/main/templates/index.md.tmpl

vaerh (changes requested):
My comments in the code.

docs/index.md
davidkern: Sure thing, I should have realized that came from the code comments...

docs/index.md
davidkern: Thanks. Out of curiosity where does the port get set for https? I was going by: https://github.com/terraform-routeros/terraform-provider-routeros/blob/9124721271cca2e9dc41c027c72f18b3b458f620/routeros/mikrotik_client.go#L84

vaerh: There is no secret, the port is embedded in the schema itself. For REST, a standard http client is used, which parses the passed URL and uses a default port (https = 443).

docs/index.md
davidkern: I've left this one alone in the revision. I don't think it is confusing since the provider docs now call out the options (and it'd be good to guide toward using the new rest api by default).

docs/index.md
davidkern: Moved the scheme documentation over to the template and included the https scheme.

vaerh (approved):

vaerh:
The documentation will automatically change after the next release.

PR #258, opened 2023-09-18
Create development environment based on libvirt/qemu

jlpedrosa:
This PR enabled the creation of a CloudHostedRouter to enable localdevelopment against virtual devicces.
It relies on terraform module "dmacvicar/libvirt" which interfaces with qemu/libvirt.

by rurring `terrraform apply` on devenv folder it will download the mikrotik routeros cloud router images and start the VM so the tests can be run against in a local fashion. 

I created this machine just to support my local development, feel free to discard it if you don't find it useful or you have a better way (if you have, please do share!).

At the end of terraform apply (~60 secs) a full router is created with 4 interfaces:

image

### DIFF
--- a/devenv/virtual_router.tf
+++ b/devenv/virtual_router.tf
@@ -1,0 +1,153 @@
+
+locals {
+  iso_dir = "${path.root}/iso"
+}
+
+
+resource "libvirt_network" "chr_network" {
+  # the name used by libvirt
+  name = "chr_network"
+
+  # mode can be: "nat" (default), "none", "route", "open", "bridge"
+  mode = "nat"
+
+  #  the domain used by the DNS server in this network
+  domain = "routerosdev.local"
+
+  #  list of subnets the addresses allowed for domains connected
+  # also derived to define the host addresses
+  # also derived to define the addresses served by the DHCP server
+  addresses = ["10.10.10.0/24", "2001:db8:ca2:2::/64"]
+
+  # (optional) the bridge device defines the name of a bridge device
+  # which will be used to construct the virtual network.
+  # (only necessary in "bridge" mode)
+  # bridge = "br7"
+
+  # (optional) the MTU for the network. If not supplied, the underlying device's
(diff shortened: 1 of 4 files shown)


devenv/virtual_router.tf
```diff
@@ -0,0 +1,153 @@
...
+}
+
+resource "libvirt_pool" "chr_image_pool" {
+  name = "chr_image_pool"
+  type = "dir"
+  path = local.iso_dir
+
+
+
+  provisioner "local-exec" {
```
jlpedrosa, line 96: this `local-exec` is necessary because the zip format and there's no download available for .iso images. Not happy, but works.

vaerh:
Jose, why don't you use the docker image for testing? All automatic tests in the repository are done within that image.

jlpedrosa:
I was going to use docker compose, but the tun interface is killing me because I can't wait for container to be ready and/or put the initializer in a container:

```
services:
  routeros:
    image: vaerhme/routeros
    ports:
      - 443:443
      - 8728:8728
      - 8729:8729
    volumes:
      - /dev/net/tun:/dev/net/tun
    cap_add:
      - NET_ADMIN
    entrypoint: /routeros/entrypoint_with_four_interfaces.sh
    healthcheck:
      test: [ "CMD", "curl", "-k", "https://localhost" ]
      interval: 1m30s
      timeout: 10s
      retries: 3
      start_period: 20s
      start_interval: 5s

  initializer:
    image: routeros-initializer
    build:
      context: .
      dockerfile: Dockerfile
    environment:
      ROS_USERNAME: admin
      ROS_PASSWORD: ''
      ROS_IP_ADDRESS: routeros
    depends_on:
      routeros:
        condition: service_healthy
```

vaerh:
Yes, there is such a problem :) But this is the module author's way and I don't want to change it without him.
P.S. I am probably not subject to this expectation, as I have the container existing in the background for a long time.

PR #260, opened 2023-09-20
Ip firewall connection tracking

jlpedrosa:
Resource to configure Ip connection tracking.


jlpedrosa:
I see CI is failing, it seems one of the properties is not available in older versions. I see two routes:
1) Eliminate the options until you decide to "unsupport" older versions
2) Create a dynamic validation based on the version of the OS running.

Let me know what route you want me to take (or other?) and/or any advice if you want route 2.

vaerh:
This is a long-standing problem. Solution in [resource_snmp_test.go](https://github.com/terraform-routeros/terraform-provider-routeros/blob/8cee7e564c3ee4695cd35a2e915fe07a2d9b6b39/routeros/resource_snmp_test.go#L15)

routeros/provider_schema_helpers.go
vaerh, line 1: I need to do some checks regarding this code

jlpedrosa: Sure! One of the reasons I did this is because the documentation for that part of the system was very flaky, some fields were not even in the documentation, missing descriptions. I felt it was sanier not to override things. 
Personally I think is a sane behaviour for a most of the fields. It also prevents drifting for values that are not user provided.

vaerh (approved):

PR #261, opened 2023-09-20
Implement routeros_system_logging resource

jlpedrosa:
This enables configuring the rules of logging for RouterOS

### DIFF
--- a/routeros/resource_system_logging.go
+++ b/routeros/resource_system_logging.go
@@ -1,0 +1,73 @@
+package routeros
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+/* {
+    ".id": "*4",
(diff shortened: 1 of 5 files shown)


routeros/resource_system_logging.go
```diff
@@ -0,0 +1,73 @@
...
+    ".id": "*4",
+    "action": "echo",
+    "default": "true",
+    "disabled": "false",
+    "invalid": "false",
+    "prefix": "",
+    "topics": "critical"
+} */
+
+var validTopics = []string{
```
vaerh, line 18: Were you planning to do validation?

jlpedrosa: My bad: I implemented validation but terraform does not support validation for lists :( and I only found out after it was implemented (error message on the sdk), I deleted the function but not the slice. 

I've just deleted the slice.

jlpedrosa: Actually I found a way to add the validation.

vaerh (approved):

PR #263, opened 2023-09-21
Create test helper function to test resource exists

jlpedrosa:
All tests included an resource specific function to test wether the resource exists in it's primary state and had ID assigned.

This PR extracts the function in a generic way and refactor it's usages across the tests.


vaerh (approved):
That's great! Thank you!

Issue #273, opened 2023-10-04
routeros_ip_address documentation example doesn't match with behavior


ArGJolan:
**What documentation needs updating?:**
In the documentation for the resource `routeros_ip_address`, the fields `address` and `network` are swapped.

RouterOS UI:
<img width="446" alt="image" src="https://github.com/terraform-routeros/terraform-provider-routeros/assets/11016546/32118684-c0fc-44ca-9f9f-418fe97e9e3a">

Documentation:
```hcl
resource "routeros_ip_address" "address" {
  address   = "10.0.0.1"
  interface = "bridge"
  network   = "10.0.0.0/24"
}
```

**What needs to be added to the documentation?:**
Swap the `address` and `network` fields.

PR #274, opened 2023-10-04
Update routeros_ip_address doc example usage to reflect actual behavior

ArGJolan:
Fixes https://github.com/terraform-routeros/terraform-provider-routeros/issues/273


vaerh (changes requested):
Thanks for the PR!
Please swap out the last octet.
address = "10.0.0.1/24"
network = "10.0.0.0"

PR #279, opened 2023-10-24
fix: Fix enumerated values in CAPsMAN resources

dokmic:
This PR fixes the following:
- add some missing values in the `routeros_capsman_channel` resource;
- fix `routeros_capsman_manager .upgrade_policy` to accept `suggest-same-version` instead of `suggest-same-upgrade`.

### DIFF
--- a/routeros/resource_capsman_channel.go
+++ b/routeros/resource_capsman_channel.go
@@ -32,8 +32,8 @@ func ResourceCapsManChannel() *schema.Resource {
 			Type:        schema.TypeString,
 			Optional:    true,
 			Description: "Define operational radio frequency band and mode taken from hardware capability of wireless card.",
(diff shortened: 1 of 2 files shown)


vaerh:
What version of ROS are you using? 
If it is the latest ROS version, then the patch will have to wait, as the functionality to select the provider scheme depending on the ROS version is under development.

dokmic:
@vaerh I don't think it matters for this particular issue. Those values had been present in 6. x, as I used them back then.
- `routeros_capsman_channel.band` -- the added values are missing in their older documentation but were there from the beginning.
- `routeros_capsman_channel. control_channel_width` -- the added values are valid for 802.11ac, which was there for ages too.
- `routeros_capsman_manager.upgrade_policy` -- there is a typo in the official [docs](https://help.mikrotik.com/docs/pages/viewpage.action?pageId=53444648) for `6.x` (see the correct one in the description column).

Either way, according to the current implementation, if a user attempts to set some of the values not supported by their ROS version, an error will be returned and nothing more.

vaerh (approved):

PR #283, opened 2023-10-30
feat: Add EoIP tunnel support

dokmic:
This PR adds Ethernet over IP tunnel support. Most of the properties have been reused, just moved them to the helpers to follow DRY.

The schema has not been changed in `7.x`, so it works in both `6.x` and `7.x` versions.


vaerh (approved):
Great work thanks a lot!

vaerh:
@dokmic 
I want to change the resource key from **Name** to **ID** in the next release, because resource names should be avoided in the future due to problems with the provider #192.
Can you test the resource on your data with this change?
``` go
		MetaResourcePath: PropResourcePath("/interface/eoip"),
		MetaId:           PropId(Id),
```

PR #287, opened 2023-11-02
Vaerh/issue268

vaerh:


routeros/resource_interface_eoip.go
```diff
@@ -9,7 +9,7 @@ import (
 func ResourceInterfaceEoip() *schema.Resource {
 	resSchema := map[string]*schema.Schema{
 		MetaResourcePath: PropResourcePath("/interface/eoip"),
-		MetaId:           PropId(Name),
+		MetaId:           PropId(Id),
```
dokmic, line 12: @vaerh I tested this particular change. Functionally, that works as expected, but it breaks the existing state. The `id` property is outdated in the state, and hence terraform loses the link. That can be easily fixed by reimporting the resource, which will update the `id` property only.

That's kind of a breaking change, but we just released this feature, so I am not sure if that's critical.

vaerh: Ok, let's make it pretty. I have added state migration, please test it. Maybe there are some other requests.

dokmic (approved):
Tested schema version bump of all the changed resources as I had all of them in my configuration. State migrated seamlessly -- I've got the correct ID's in the state instead of the resource names.

Great job! LGTM 👍

PR #299, opened 2023-11-12
fix: Skip computed stat fields

### LINKED ISSUE #298: Schema warnings when managing Ethernet Interfaces
**Describe the bug**

A lot of warning appear in the console when trying to manage `routeros_interface_ethernet` objects

**To Reproduce**

Using provider version `1.23.0` and RouterOS version `7.9.2`, try to create a new `routeros_interface_ethernet` object.

**Expected behavior**

The object should be created without generating any warnings or errors.

**Stack Trace**

Applying the following config:

```hcl
terraform {
  required_providers {
    routeros = {
      source  = "terraform-routeros/routeros"
      version = "1.23.0"
    }
  }
}

provider "routeros" {
  hosturl  = var.mikrotik_router_url
  username = var.mikrotik_router_username
  password = var.mikrotik_router_password
  insecure = var.mikrotik_router_insecure_skip_tls_verify
}

resource "routeros_interface_ethernet" "desktop-iface" {
  factory_name = "sfp-sfpplus1"
  name         = "desktop"
  l2mtu        = 1514
}

```

Produces this result:

```bash
mike@DESKTOP-48UPPAD:/com.docker.devenvironments.code/infra/mikrotik$ terraform apply

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

  # routeros_interface_ethernet.desktop-iface will be created
  + resource "routeros_interface_ethernet" "desktop-iface" {
... (ommited for brevity)
    }

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

routeros_interface_ethernet.desktop-iface: Creating...
routeros_interface_ethernet.desktop-iface: Creation complete after 0s [id=*9]
╷
│ Warning: Field 'tx_deferred' not found in the schema
│ 
│   with routeros_interface_ethernet.desktop-iface,
│   on interfaces.tf line 1, in resource "routeros_interface_ethernet" "desktop-iface":
│    1: resource "routeros_interface_ethernet" "desktop-iface" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'tx_deferred': '0' ◁
╵
╷
│ Warning: Field 'tx_excessive_collision' not found in the schema
│ 
│   with routeros_interface_ethernet.desktop-iface,
│   on interfaces.tf line 1, in resource "routeros_interface_ethernet" "desktop-iface":
│    1: resource "routeros_interface_ethernet" "desktop-iface" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'tx_excessive_collision': '0' ◁
╵
╷
│ Warning: Field 'tx_single_collision' not found in the schema
│ 
│   with routeros_interface_ethernet.desktop-iface,
│   on interfaces.tf line 1, in resource "routeros_interface_ethernet" "desktop-iface":
│    1: resource "routeros_interface_ethernet" "desktop-iface" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'tx_single_collision': '0' ◁
╵
╷
│ Warning: Field 'tx_fcs_error' not found in the schema
│ 
│   with routeros_interface_ethernet.desktop-iface,
│   on interfaces.tf line 1, in resource "routeros_interface_ethernet" "desktop-iface":
│    1: resource "routeros_interface_ethernet" "desktop-iface" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'tx_fcs_error': '0' ◁
╵
╷
│ Warning: Field 'tx_multiple_collision' not found in the schema
│ 
│   with routeros_interface_ethernet.desktop-iface,
│   on interfaces.tf line 1, in resource "routeros_interface_ethernet" "desktop-iface":
│    1: resource "routeros_interface_ethernet" "desktop-iface" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'tx_multiple_collision': '0' ◁
╵

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

**Additional context**

When using ROS v7.12 (stable), this part outright fails with an error saying that `full_duplex` is an unknown argument. I had to downgrade my ROS version.

dokmic:
While working with the provider, I discovered that it changes the Terraform state on every apply, even though there are no changes. That's caused by constantly changing stats fields (e.g., number of transferred bytes or packets). That breaks the whole concept of the Terraform state and produces state changes without actual "infrastructure" change.

Apart from that, some resources returned many warnings related to the missing schema fields.

This PR tries to address those problems in a backward-compatible way by reusing the `MetaSkipFields` property:
1. Skip stats for the ethernet resources, which were causing a lot of noise in the output (resolves #298).
2. Skip stats for the firewall rules.
3. Skip debug fields from the bridge ports.
4. Skip the last login information for the system users.
5. Add the missing computed field to the system logging resource.

As this PR doesn't explicitly add any new schema fields, the changes are compatible with any supported RouterOS version.

### DIFF
--- a/routeros/resource_interface_bridge_port.go
+++ b/routeros/resource_interface_bridge_port.go
@@ -71,6 +71,7 @@ func ResourceInterfaceBridgePort() *schema.Resource {
 	resSchema := map[string]*schema.Schema{
 		MetaResourcePath: PropResourcePath("/interface/bridge/port"),
 		MetaId:           PropId(Id),
+		MetaSkipFields:   PropSkipFields(`"debug_info","port_number"`),
 
 		"nextid": {
 			Type:     schema.TypeString,
@@ -103,10 +104,6 @@ func ResourceInterfaceBridgePort() *schema.Resource {
 				"When disabled, drops broadcast traffic on egress ports. ",
 		},
 		KeyComment: PropCommentRw,
-		"debug_info": {
-			Type:     schema.TypeString,
-			Computed: true,
-		},
 		"designated_bridge": {
 			Type:        schema.TypeString,
 			Computed:    true,
@@ -254,12 +251,6 @@ func ResourceInterfaceBridgePort() *schema.Resource {
 			Type:        schema.TypeBool,
 			Computed:    true,
 			Description: "Whether the port is connected to a bridge port using full-duplex (true) or half-duplex (false).",
-		},
-		"port_number": {
-			Type:     schema.TypeInt,
-			Computed: true,
-			Description: "Port number will be assigned in the order that ports got added to the bridge, " +
-				"but this is only true until reboot. After reboot internal numbering will be used.",
 		},
 		"priority": {
 			Type:     schema.TypeString,
--- a/routeros/resource_interface_ethernet.go
+++ b/routeros/resource_interface_ethernet.go
@@ -54,7 +54,16 @@ func ResourceInterfaceEthernet() *schema.Resource {
 	resSchema := map[string]*schema.Schema{
 		MetaResourcePath: PropResourcePath("/interface/ethernet"),
 		MetaId:           PropId(Id),
-		MetaSkipFields:   PropSkipFields(`"factory_name"`),
+		MetaSkipFields:   PropSkipFields(`"factory_name","driver_rx_byte","driver_rx_packet","driver_tx_byte","driver_tx_packet",` +
+			`"rx_64","rx_65_127","rx_128_255","rx_256_511","rx_512_1023","rx_1024_1518","rx_1519_max",` +
+			`"tx_64","tx_65_127","tx_128_255","tx_256_511","tx_512_1023","tx_1024_1518","tx_1519_max",` +
+			`"tx_rx_64","tx_rx_65_127","tx_rx_128_255","tx_rx_256_511","tx_rx_512_1023","tx_rx_1024_1518","tx_rx_1519_max",` +
+			`"rx_broadcast","rx_bytes","rx_control","rx_drop","rx_fcs_error","rx_fragment","rx_jabber","rx_multicast","rx_packet","rx_pause","rx_too_short","rx_too_long",` +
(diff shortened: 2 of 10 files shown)


vaerh:
There's a big problem: someone can use some of these fields and we'll get a script that doesn't work.
If you have relevant experience, I wouldn't mind an answer to the question why changing the state file breaks the terraforming idea? I've looked at the code of other providers but I can't find how they handle changing fields.
Somewhere I'm thinking that the counters should probably be moved to the data sources, but that's unrealistic.

dokmic:
@vaerh Changing state breaks the main principle of IaC -- [idempotence](https://learn.microsoft.com/en-us/devops/deliver/what-is-infrastructure-as-code#avoid-manual-configuration-to-enforce-consistency).
> Idempotence, the ability of a given operation to always produce the same result, is a crucial IaC principle. A deployment command always sets the target environment into the same configuration, regardless of the environment's starting state.

In our case, the state was constantly changing, requiring reuploading the state to the selected backend. This is very inefficient and may lead to extra costs and make debugging more difficult.

I think there is no practical or theoretical use case when you need counters in our infrastructure. The resource fields are usually used as dependencies in other resources or outputs, and conditions based on counters will be unreliable. It would make sense to have them in the data sources, but again, that shouldn't be used in any conditions. And the whole idea of having a script (which is imperative) is against the [declarative](https://developer.hashicorp.com/terraform/language#about-the-terraform-language) model of Terraform language:
> The Terraform language is declarative, describing an intended goal rather than the steps to reach that goal.

On top of that, having more defined schema fields requires a little bit more maintenance as it's harder to predict them and, therefore, test them. So, pragmatically, excluding them from the resource definitions is better. In the worst case, people using those fields will come back here after bumping the provider.

vaerh:
Thanks!
I thought that the feedback of the system in the form of a state file does not affect idempotence.
It is likely that we will have to remove all counters from resources.
I need some time to think about this.

vaerh (approved):

PR #356, opened 2024-02-13
feat(container): Add support for /container

ksmets:
Add support for:
- /container/config
- /container/envs
- /container/mounts
- /container

No tests provided, installation of container package works, but can't enable container mode.


routeros/mikrotik_crud.go
```diff
@@ -97,7 +97,7 @@ func UpdateItem(id *ItemId, resourcePath string, item MikrotikItem, c Client) (M
 
 func DeleteItem(id *ItemId, resourcePath string, c Client) error {
 	if id.Value == "" {
-		return errEmptyId
+		return errEmptyPath
```
vaerh, line 100: Is that a misprint?

ksmets: Yes, good catch. Reverted error mix up.

PR #357, opened 2024-02-13
fix(ipv6): Add fields for /ipv6/dhcp-client

ksmets:


routeros/resource_ipv6_dhcp_client.go
```diff
@@ -0,0 +1,138 @@
...
+		KeyComment: PropCommentRw,
+		"default_route_distance": {
+			Type:         schema.TypeInt,
+			Optional:     true,
+			Default:      1,
+			Description:  "Distance of default route. Applicable if add-default-route is set to yes.",
+			ValidateFunc: validation.IntBetween(0, 255),
+			// Default route distance returns as empty when the dhcp-client is searching.
+			// This produces inconsistent results, for this case, we will suppress changes.
+			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
vaerh, line 34: I don't have an IPv6 network at hand, could you test this field with ```Default``` excluded and ```DiffSuppressFunc: AlwaysPresentNotUserProvided```? Will the desired schema behavior be present in this combination?

vaerh (changes requested):

routeros/resource_ipv6_dhcp_client.go
ksmets: Seems to work with your requested changes. I just used the same solution as already present in ResourceDhcpClient for /ip/dhcp-client.

vaerh: Now there is a small mix of different realizations of the same fucntional. Michael Dokolin suggested to use this variant for all parameters returning values, so that it would be easier to ensure compatibility of different versions of resource schemas in the future.

ksmets: I see, will adapt.

vaerh (approved):

vaerh:
Oh yes it's a busy day, I apologize for my inattentiveness. Jose Luis Pedrosa has already added this resource. I'll have to get both of your schemas into the same view.
:man_facepalming:

ksmets:
No worries @vaerh, I took the liberty to merge both schemas.

jlpedrosa:
hey! thanks for the fixes @ksmets  There are a bunch of fields that you added that were not documented. I was doing a PR to add them!
One note: `dhcp_options` should be a list? @vaerh  any thoughts?

vaerh:
@jlpedrosa 
Probably a list is better for ease of configuration. It roughly looks like this:
``` go
		"dhcp_option": {
			Type: schema.TypeSet,
			Optional: true,
			Element: &schema.Schema{
				Type: schema.TypeString,
				// ??? if needed: ValidateFunc: validation.StringInSlice([]string{"", ""}, false,) / ValidationMultiValInSlice 
			},
			Description: "Options that are sent to the DHCP server.",
			// ??? if needed: DiffSuppressFunc: AlwaysPresentNotUserProvided,
		},
```

ksmets:
@vaerh @jlpedrosa. To me it seems that dhcp_options should be a list string and that references the names of the resources managed by `/ipv6/dhcp-client/options` resources just like in the IPv4 case. However, there aren't any predefined ones like `clientid`, `client_duid`, `hostname` as in IPv4. I will add the dhcp_options field to this PR and one to the manage the options in a separate one.

vaerh (approved):

vaerh:
Great, let's do the release!

PR #358, opened 2024-02-13
fix(interface): Add gateway6 field to /interface/veth

ksmets:


routeros/resource_interface_veth.go
```diff
@@ -33,9 +33,19 @@ func ResourceInterfaceVeth() *schema.Resource {
 		KeyDisabled: PropDisabledRw,
 		"gateway": {
 			Type:         schema.TypeString,
-			Required:     true,
+			Optional:     true,
+			Computed:     true,
```
jlpedrosa, line 37: In winbox, I can set this field. Are you sure it's computed?

ksmets: You're right, it should only be optional, not computed.

vaerh (approved):

PR #365, opened 2024-02-26
Fix max value for MTU - Fix Main

jlpedrosa:
Last PR broke main build avoiding the succesful release of the new version.


vaerh (approved):
Yes, what a beautiful solution! Thank you!

Issue #368, opened 2024-03-02
The release 1.35.1 was not generated.


jlpedrosa:
**Which pipeline fails?:**
The release 1.35.1 was not generated. The pipeline for release did not run.
I have the hunch that probably [here](https://github.com/terraform-routeros/terraform-provider-routeros/blob/main/.github/workflows/release.yml#L23)  the comparision should be `==` instead of `!=` ? 

I also see that for the non release commits, the pipeline ran anyway.

**What job does the pipeline fail on?:**
Pipeline skipped.

![image](https://github.com/terraform-routeros/terraform-provider-routeros/assets/4932947/fe1cf305-4740-417c-a349-3043988134b2)

vaerh:
Yes, there is such a problem. I'll post the artifacts manually soon and see where the build is stuck.

vaerh:
To my regret I can't manually add artifacts, as I don't have some of the necessary parameters for that. 

The release problem is old and is related to overflow in x32 architectures. The idea to fix it is beautiful: `validation.IntBetween(0, math.MaxInt32)`, but it seems to me that we need to remove this validator. It will not work properly in either x32 or x64 architecture (`MaxInt32  = 1<<31 - 1           // 2147483647`). I see two options (to comply with the MT specification):
* remove the validator
* remove the validator and make the parameter a string (to make it work correctly in x32).
I am leaning towards the first option. It is not clear how many people use the provider and this resource on the architecture x32, but the rest I do not see the point of complicating life.

jlpedrosa:
Humm... that is strange. MaxInt32 specify the bit length, so it should work on both. I agree, I don't think that validator is really important. We can also add a very large (int the context of MTU) and call it a day. What ever you think is best.

BTW, is there any way to have a sync conversation @vaerh ?

jlpedrosa:
https://github.com/terraform-routeros/terraform-provider-routeros/pull/380

jlpedrosa:
@vaerh  I saw you added more tags, but still there are no new releases on the terraform page. the build is skipped.

are you sure this code block is correct?
```
jobs:

  release:
    if: contains(github.event.head_commit.message, 'chore(release)') != true
 ```
 
 It feels like the when the commit message contains 'chore(release)' the release pipeline should be executed no?

vaerh:
@jlpedrosa 
Everything is working as expected. The artifacts are in place and should therefore be available in the HashiCorp registry.
[...]

PR #369, opened 2024-03-02
fix: release pipeline should be executed on releases only

jlpedrosa:
I think the pipeline should be executed on the positve case of the condition.

Fixes: https://github.com/terraform-routeros/terraform-provider-routeros/issues/368

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,7 @@ on:
 jobs:
 
   release:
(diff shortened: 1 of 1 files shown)


vaerh:
Hi, I'm back. Everything has been written correctly, I don't see much point in doing ordinary releases when creating a PR dependabot. The maximum is to update dependencies when new security patches are released, but this can be solved within the PR description.

PR #372, opened 2024-03-06
feat: Add `routeros_wifi` resource to manage WiFi interfaces

dokmic:
The goal of this PR is to add support for managing WiFi interfaces, but that required some additional changes:
- Refactored `AlwaysPresentNotUserProvided` helper to handle map types correctly.
- Moved L2 MTU property into a reusable helper.
- Fixed `routeros_wifi_configuration` to align with the newly introduced `routeros_wifi`.

### DIFF
--- a/routeros/provider_schema_helpers.go
+++ b/routeros/provider_schema_helpers.go
@@ -556,28 +556,6 @@ var (
 	}
 )
 
-func isRawValueEmpty(value cty.Value, path string) bool {
-	var attr cty.Value
-	keys := strings.SplitN(path, ".", 2)
-	key := keys[0]
-
-	if key == "#" || key == "%" {
-		return false
-	}
-
-	switch {
-	case value.Type().IsObjectType():
-		attr = value.GetAttr(key)
-	case value.Type().IsMapType():
-		attr = value.Index(cty.StringVal(key))
-	// Lists and sets should not be walked down as they are always updated as a whole.
-	default:
-		return false
-	}
-
-	return attr.IsNull() || len(keys) > 1 && isRawValueEmpty(attr, keys[1])
-}
-
 // Properties DiffSuppressFunc.
 var (
 	TimeEquall = func(k, old, new string, d *schema.ResourceData) bool {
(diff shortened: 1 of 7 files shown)


routeros/provider_schema_helpers.go
```diff
@@ -334,6 +334,14 @@ var (
 		Description: "Layer2 Maximum transmission unit. " +
 			"[See](https://wiki.mikrotik.com/wiki/Maximum_Transmission_Unit_on_RouterBoards).",
 	}
+	PropL2MtuRw = &schema.Schema{
+		Type:     schema.TypeInt,
+		Optional: true,
+		Description: "Layer2 Maximum transmission unit. " +
+			"[See](https://wiki.mikrotik.com/wiki/Maximum_Transmission_Unit_on_RouterBoards).",
+		ValidateFunc:     validation.IntBetween(0, 65536),
```
vaerh, line 342: Michael, greetings!
The range should be, I think, from 1 to 65535
```
		if v < min || v > max {
			errors = append(errors, fmt.Errorf("expected %s to be in the range (%d - %d), got %d", k, min, max, v)))
			return warnings, errors
		}
```

routeros/provider_schema_helpers.go
```diff
@@ -606,12 +636,7 @@ var (
 	AlwaysPresentNotUserProvided = func(k, old, new string, d *schema.ResourceData) bool {
 		// For lists and sets, the key will look like `something.12345` or `something.#`.
 		// But in the raw config it will be just `something`.
-		k = strings.Split(k, ".")[0]
-
-		if old != "" && d.GetRawConfig().GetAttr(k).IsNull() {
-			return true
-		}
-		return false
+		return old != "" && isRawValueEmpty(d.GetRawConfig(), k)
```
vaerh, line 639: Maybe it makes sense not to make the check a separate function, since it is unlikely to be reused? And the logic of the decision will be located in one place.

vaerh (changes requested):

vaerh:
@dokmic Can you give feedback on my review?

routeros/provider_schema_helpers.go
dokmic: In the documentation, it [says](https://help.mikrotik.com/docs/display/ROS/Ethernet#Ethernet-Properties) `[0..65536]`. This seems like inclusive 0 and 65536, but that doesn't make any sense. 

Anyway, they have their own validation per each resource, and, I guess, also device-specific. So, I've updated my PR.

routeros/provider_schema_helpers.go
dokmic: Well, my intent was to make it readable with the tailing recursion, but the helper itself seems a bit useless now.

I've refactored the helper to do the same without recursion as we get another set of parameters. It should still be readable 😀

dokmic:
@vaerh Thanks a lot for your feedback - I've updated my PR accordingly. Could you please take another look?

vaerh (approved):

vaerh:
Great, everything seems fine!

PR #380, opened 2024-03-17
fix: stop using Int32MaxValue as upper bound for validation

jlpedrosa:
Set 90000 (10x JumboFrame) for the upper bound of the MTU.

### DIFF
--- a/routeros/resource_ipv6_neighbor_discovery.go
+++ b/routeros/resource_ipv6_neighbor_discovery.go
@@ -1,8 +1,6 @@
 package routeros
 
 import (
-	"math"
-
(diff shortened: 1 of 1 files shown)


vaerh:
I'm gonna go look at the resource documentation. I don't understand what MT suggests as the upper dimension for MTU.

vaerh:
The maximum [MTU](https://help.mikrotik.com/docs/display/ROS/MTU+in+RouterOS#MTUinRouterOS-MAC/Layer-2/L2MTU) supported MT (from what I found) is 10226. I was wrong about MaxInt32 not working. Let it stay 10 x Jumbo for such a weird spec it's not bad.

vaerh (approved):

PR #359, opened 2024-02-14
Add missing ethernet interface fields to SkipFields

jlpedrosa:
On latest version of Router OS: 7.13.4 The following warnings are pressent for interfaces of type ethernet:

```
│ Warning: Field 'tx_rx_1024_max' not found in the schema
│ 
│   with routeros_interface_ethernet.eths["core-link-2"],
│   on interfaces.tf line 87, in resource "routeros_interface_ethernet" "eths":
│   87: resource "routeros_interface_ethernet" "eths" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'tx_rx_1024_max': '2557838' ◁

│ Warning: Field 'tx_unicast' not found in the schema
│ 
│   with routeros_interface_ethernet.eths["core-link-2"],
│   on interfaces.tf line 87, in resource "routeros_interface_ethernet" "eths":
│   87: resource "routeros_interface_ethernet" "eths" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'tx_unicast': '3940625' ◁

│ Warning: Field 'rx_error_events' not found in the schema
│ 
│   with routeros_interface_ethernet.eths["core-link-2"],
│   on interfaces.tf line 87, in resource "routeros_interface_ethernet" "eths":
│   87: resource "routeros_interface_ethernet" "eths" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'rx_error_events': '0' ◁
│ Warning: Field 'rx_unicast' not found in the schema
│ 
│   with routeros_interface_ethernet.eths["core-link-2"],
│   on interfaces.tf line 87, in resource "routeros_interface_ethernet" "eths":
│   87: resource "routeros_interface_ethernet" "eths" {
│ 
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'rx_unicast': '1406391' ◁
│ 
```

As the fields are read only, adding them to skip:
- tx_rx_1024_max
- tx_unicast
- rx_error_events
- rx_unicast

PR #381, opened 2024-03-18
feat: Resolve warnings due to some missing properties

dokmic:
This PR adds/fixes the following:
- Skip the newly introduced `debug_info` property in the `routeros_interface_bridge_vlan` resource - added in 7.14.
- Add the newly introduced `push_routes` property to the `routeros_ovpn_server` resource - added in 7.14.
- Fix corrupted skipped fields meta property in `routeros_interface_ethernet` - regression from #359.

### DIFF
--- a/routeros/resource_interface_ethernet.go
+++ b/routeros/resource_interface_ethernet.go
@@ -56,10 +56,10 @@ func ResourceInterfaceEthernet() *schema.Resource {
 			`"factory_name","driver_rx_byte","driver_rx_packet","driver_tx_byte","driver_tx_packet",` +
 				`"rx_64","rx_65_127","rx_128_255","rx_256_511","rx_512_1023","rx_1024_1518","rx_1519_max",` +
 				`"tx_64","tx_65_127","tx_128_255","tx_256_511","tx_512_1023","tx_1024_1518","tx_1519_max",` +
-				`"tx_rx_64","tx_rx_65_127","tx_rx_128_255","tx_rx_256_511","tx_rx_512_1023","tx_rx_1024_1518","tx_rx_1024_max",tx_rx_1519_max",` +
+				`"tx_rx_64","tx_rx_65_127","tx_rx_128_255","tx_rx_256_511","tx_rx_512_1023","tx_rx_1024_1518","tx_rx_1024_max","tx_rx_1519_max",` +
 				`"rx_broadcast","rx_bytes","rx_control","rx_drop","rx_fcs_error","rx_fragment","rx_jabber","rx_multicast","rx_packet","rx_pause","rx_too_short","rx_too_long",` +
 				`"tx_broadcast","tx_bytes","tx_control","tx_drop","tx_fcs_error","tx_fragment","tx_jabber","tx_multicast","tx_packet","tx_pause","tx_too_short","tx_too_long",` +
-				`"rx_align_error","rx_carrier_error","rx_code_error","rx_error_events","rx_length_error","rx_overflow","rx_unicast","rx_unknown_op"` +
+				`"rx_align_error","rx_carrier_error","rx_code_error","rx_error_events","rx_length_error","rx_overflow","rx_unicast","rx_unknown_op",` +
 				`"tx_collision","tx_excessive_collision","tx_late_collision","tx_multiple_collision","tx_single_collision","tx_total_collision",` +
 				`"tx_deferred","tx_excessive_deferred","tx_unicast","tx_underrun"`,
 		),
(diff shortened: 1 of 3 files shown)


routeros/resource_ovpn_server.go
vaerh, line 1: Michael, please change the attribute type to `schema.TypeSet`. I've been trying to use it for lists lately, as there are problems when returning sorted values from MT for an unsorted plan. The plan becomes inconsistent.

vaerh (changes requested):

routeros/resource_interface_ethernet.go
```diff
@@ -56,10 +56,10 @@ func ResourceInterfaceEthernet() *schema.Resource {
 			`"factory_name","driver_rx_byte","driver_rx_packet","driver_tx_byte","driver_tx_packet",` +
```
jlpedrosa, line 56: Should we create a function `PropSkipFieldsSlice(items ...string)` So we can actually be sure that this not happen in the future? 
What do you guys think? @vaerh ?

jlpedrosa: I mean something like: https://github.com/terraform-routeros/terraform-provider-routeros/pull/382

routeros/resource_ovpn_server.go
dokmic: It makes sense in this particular case. I will keep that in mind and try fixing it when the list is actually needed. Thanks 👍

routeros/resource_interface_ethernet.go
dokmic: @jlpedrosa I would rather change the existing `PropSkipFields` and replace all the occurrences to pass an array instead of a string. I also wanted to do that a while ago.

vaerh: I think it makes sense. But it should be done for all meta-attributes at once.
I changed the function a bit to make it work a bit faster:
``` go
func BenchmarkSlice1(b *testing.B) {
	for i := 0; i < b.N; i++ {
		PropSkipFieldsSlice1("0", "1","2", "3","4", "5","6", "7","8","9")
	}
}

func BenchmarkSlice2(b *testing.B) {
	for i := 0; i < b.N; i++ {
		PropSkipFieldsSlice2("0", "1","2", "3","4", "5","6", "7","8","9")
	}
}

func PropSkipFieldsSlice1(s ...string) string {
	builder := strings.Builder{}
	for _, item := range s {
		builder.WriteString(fmt.Sprintf("\"%s\",", item))
	}
	res := builder.String()
	lastComaRemoved := res[0 : len(res)-1]
	return lastComaRemoved
}

func PropSkipFieldsSlice2(s ...string) string {
	res := make([]string, len(s))
	for i, item := range s {
		res[i] = `"`+item+`"`
	}
	return strings.Join(res,",")
}
```
``` bash
BenchmarkSlice1
BenchmarkSlice1-8         671860              1533 ns/op             320 B/op         24 allocs/op
BenchmarkSlice2
BenchmarkSlice2-8        1836508               611.9 ns/op           240 B/op         12 allocs/op
```

``` go
func stringFromSlice(s []string) string {
	res := make([]string, len(s))
	for i, item := range s {
		res[i] = `"`+item+`"`
	}
	return strings.Join(res,",")
}

func PropSkipFields(s ...string) *schema.Schema {
	return &schema.Schema{
		Optional:    true,
		Default:     stringFromSlice(s),
		Description: "<em>A set of transformations for field names. This is an internal service field, setting a value is not required.</em>",
		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
			return true
		},
	}
}
```

vaerh (approved):

routeros/resource_interface_ethernet.go
vaerh: I suggest we split this PR. We accept this one as it is.
Now I would like to start the process of publishing relises in the Hashi registry. There are problems with this.

jlpedrosa: Actually, most likely the "heaviness" of my implementation comes from the sprintf, probably if you copy my loop with your string concat is going to be faster . Strings.join loops over the slice again. 100% up to you I don't think this is very performance sensitive.

vaerh: You're right. It's Sprintf
``` bash
BenchmarkSlice1
BenchmarkSlice1-8 2408424 477.1 ns/op 120 B/op 4 allocs/op
BenchmarkSlice2
BenchmarkSlice2-8 1857993 616.3 ns/op 240 B/op 12 allocs/op
```

dokmic: @vaerh Could we maybe merge this PR?   It should not be a problem to release it later.

PR #393, opened 2024-03-23
Implement ip/service data source

### LINKED ISSUE #391: data routeros_ip_service
I want to disable all IP services except those I have `local.services` configuration for.

With a data source for ` GET /rest/ip/service`, I could `for_each` the result and exclude the service(s) I want to configure, and easily disable them all without hardcoding a list.

**Additional context**
For other use cases, and as is typical for such a data source, it would be good if it were filterable. On VRF or disabled attrs, would be the most obviously useful.

OJFord:
Closes #391.

Used successfully like:
```tf
data "routeros_ip_services" "router" {
  provider = routeros.router
}

resource "routeros_ip_service" "router-disabled" {
  provider = routeros.router
  for_each = { for s in data.routeros_ip_services.router.services : s.name => s if s.name != "www-ssl" }

  disabled = true
  numbers  = each.value.name
  port     = each.value.port
}
```

to test.


vaerh:
That's great! I have a request for you to add an example file for this data source `/examples/data-sources/...`
The release of your add-on will be after fixing the publication in the TF registry.

vaerh (changes requested):
https://github.com/terraform-routeros/terraform-provider-routeros/pull/393#issuecomment-2016580895

vaerh (approved):
I won't torture you with my requests. If you don't mind, I've added the example code from this PR as an example for the resource. To make it easier for other users.
P.S. The release build process is designed so that all the contents of the `docs` folder are created again each time. For this reason all examples should be placed in `examples`.

OJFord:
Ah sorry, no problem, I just copied one of the others that only had a basic example like that, but yes good idea.

Thanks!

PR #398, opened 2024-03-26
feat: Add `http` scheme support for the REST API

dokmic:
This PR enables the `http` scheme support in the connection settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ terraform {
 }
 
 provider "routeros" {
-  hosturl  = "(https|api|apis)://my.router.local[:port]"
+  hosturl  = "(http|https|api|apis)://my.router.local[:port]"
   username = "my_username"
   password = "my_super_secret_password"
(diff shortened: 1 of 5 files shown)


vaerh:
Hi Michael!

Originally this functionality was in the provider but on early versions of ROS it didn't work for some reason.

Please look at the differences between your PR and my implementation. It seems to me that `useTLS = false` is still needed.

https://github.com/vaerh/terraform-provider-routeros/commit/a7869fbc5bdfc95c725ecc1b66763969b2c9716c

vaerh (changes requested):
Please see the comment on the PR

dokmic:
@vaerh I've been exploiting this feature locally for a few weeks already, and so far so good. It works perfectly with a few ROS instances.

The `useTLS = false` statement is unnecessary, as it only matters when the selected transport is `TransportAPI`.

PR #423, opened 2024-04-15
fix(no-release): [Backward compatibility] routeros_interface_bridge NOT WORKING  as expected

### LINKED ISSUE #417: [Backward compatibility] routeros_interface_bridge NOT WORKING  as expected
```
resource "routeros_interface_bridge" "bridge" {
  name           = "bridge"
}
```
i expect  that   when. i use  name           = "bridge" 
only  name is set  on a mikrotik
but for some reason a lot of other  not  needed staff is set also 

image

vaerh:
Fixes #417


gudge25:
arp-timeout=auto  not  needed
image

PR #426, opened 2024-04-15
fix(no-release): [Backward compatibility] routeros_interface_bridge_port

### LINKED ISSUE #419: [Backward compatibility] routeros_interface_bridge_port
from example
```
resource "routeros_interface_bridge_port" "bridge_port" {
  bridge    = "bridge"
  interface = "ether2"
  pvid      = "50"
}
```

but on older  router OS there  is no 
pvid      = "50"  
parameter at all

here are "REQUIRED" fields
image

here is what is sending

image

vaerh:
Fixes #419


gudge25:
tested on  
- Installed terraform-routeros/routeros v1.46.1 (self-signed, key ID 45571F2D7311B119)
"**pvid**"  is still required  

image

PR #439, opened 2024-05-01
feat: add ability to sign certificates with scep

stasjok:
* add ability to sign certificates with scep (`add-scep` command)
* add missing computed `challenge_password` parameter (present only with scep client)
* add a couple of tests for resource updating, resource attributes are checked from the actual API
* `ForceNew` is not enough on `sign` set. It need to be set on all elements or recreation won't work

There are still some quirks I found:
1. If for the following resources:
```tf
resource "routeros_system_certificate" "root_ca" {
  name        = "Test-Root-CA"
  common_name = "RootCA"
  key_usage   = ["key-cert-sign", "crl-sign"]
  trusted     = true
  sign {
  }
}

resource "routeros_system_certificate" "server_crt" {
  name        = "Server-Certificate"
  common_name = "server.crt"
  key_usage = ["digital-signature", "key-encipherment", "tls-server"]
  sign {
    ca = routeros_system_certificate.root_ca.name
  }
}
```
user would change something in a CA (but not the name):
```tf
resource "routeros_system_certificate" "root_ca" {
  name        = "Test-Root-CA"
  common_name = "RootCA"
  key_usage   = ["key-cert-sign", "crl-sign"]
  trusted     = true
  sign {
    ca_crl_host = "10.1.1.1"
  }
}

resource "routeros_system_certificate" "server_crt" {
  name        = "Server-Certificate"
  common_name = "server.crt"
  // KUs: igitalSignature, keyEncipherment or keyAgreement
  key_usage = ["digital-signature", "key-encipherment", "tls-server"]
  sign {
    ca = routeros_system_certificate.root_ca.name
  }
}
```
then there are errors that resource not found. It's because RouterOS removes all dependent certificates with a CA and the provider can't handle it for now.

2. An ability to remove certificates was added to RouterOS 7.12 (https://mikrotik.com/download/changelogs#c-stable-v7_12: `certificate - allow to remove issued certificates when CRL is not used`), but when CA certificate doen't have crl set. I didn't find a helper to check RouterOS version in the provider. But maybe it can be implemented simpler: just try to delete certificate, and in case there is an error, that certificate can only be revoked, then revoke it. It can be useful for changing certificate attributes without changing the name (otherwise there is a name conflict).

### DIFF
--- a/examples/resources/routeros_system_certificate/resource.tf
+++ b/examples/resources/routeros_system_certificate/resource.tf
@@ -39,3 +39,13 @@ resource "routeros_system_certificate" "unsigned_crt" {
   key_size         = "1024"
   subject_alt_name = "DNS:router.lan,DNS:myrouter.lan,IP:192.168.88.1"
 }
+
+resource "routeros_system_certificate" "scep_client" {
+  name        = "SCEP-Client"
+  common_name = "scep-client.crt"
+  key_usage   = ["digital-signature", "key-agreement", "tls-client"]
+
+  sign_via_scep {
+    scep_url = "http://scep.server/scep/test"
+  }
+}
--- a/routeros/mikrotik_client.go
+++ b/routeros/mikrotik_client.go
@@ -31,6 +31,7 @@ const (
 	crudDelete
 	crudPost
 	crudSign
+	crudSignViaScep
 	crudRemove
 	crudRevoke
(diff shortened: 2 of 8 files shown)


vaerh (approved):
Yeah, it's a lot of work. Thank you so much for your contribution!

PR #451, opened 2024-05-10
Fix: suppress_hw_offload as readwrite instead of computed

danielnachtrub:


vaerh (changes requested):
Daniel, please add ```DiffSuppressFunc: AlwaysPresentNotUserProvided,``` to the attribute.

vaerh (approved):

vaerh (approved):

PR #465, opened 2024-05-26
Fix unintended plan diff from unspecified defaults

OJFord:
The default value for `auto_mac` is false, and `lease_time` 30m, but if unspecified they're assumed true and 10m respectively; so a plan with them unspecified and having imported the default configuration from a freshly reset routeros wants to change them.

### DIFF
--- a/routeros/resource_ip_dhcp_server.go
+++ b/routeros/resource_ip_dhcp_server.go
@@ -48,11 +48,11 @@ func ResourceDhcpServer() *schema.Resource {
 				"* lease-time - use time from lease-time parameter",
 		},
 		"bootp_support": {
-			Type:         schema.TypeString,
-			Optional:     true,
-			Default:      "static",
-			Description:  "Support for BOOTP clients.",
-			ValidateFunc: validation.StringInSlice([]string{"none", "static", "dynamic"}, false),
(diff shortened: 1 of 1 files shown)


vaerh (changes requested):
Please remove the `Default` values and add `DiffSuppressFunc: AlwaysPresentNotUserProvided,`. In the current context it is more correct (no problems with working under different `ROS` versions)

routeros/resource_interface_bridge.go
```diff
@@ -37,6 +37,7 @@ func ResourceInterfaceBridge() *schema.Resource {
 		"auto_mac": {
 			Type:     schema.TypeBool,
 			Optional: true,
+			Default:  false,
```
vaerh, line 40: This is proper provider behavior. The `true` value is returned exactly according to the documentation: https://wiki.mikrotik.com/wiki/Manual:Interface/Bridge.
`auto-mac (yes | no; Default: *yes*)`
So we need to revert this fix back :)

OJFord: Sorry just saw your comment, I thought I'd already removed that early but missed it in commit - just removed that one too a little while ago. Cheers.

vaerh (approved):

PR #468, opened 2024-05-27
Fix interface list member missing `comment`

OJFord:


vaerh (changes requested):
I don't mind, but please replace with `KeyComment: PropCommentRw,`.
You can keep the documentation intact, it is generated at each release and in the HC registry it is published with it.

OJFord:
Ah 👍🏻 thanks

PR #498, opened 2024-06-22
fix(snmp-community): Allow a set of `addresses`

### LINKED ISSUE #495: routeros_snmp_community Addresses Doesn't Accept Multiple Addresses
**Describe the bug**
routeros_snmp_community `addresses` doesn't allow multiple address.

**To Reproduce**
```terraform
resource "routeros_snmp_community" "mything" {
  addresses = "10.0.1.12,10.10.0.129"
  name      = "mything"
}
```

Fails validation

**Expected behavior**
```terraform
resource "routeros_snmp_community" "mything" {
  addresses = ["10.0.1.12", "10.10.0.129"]
  name      = "mything"
}
```
Should yeild 
```

  # routeros_snmp_community.mything will be updated in-place
  ~ resource "routeros_snmp_community" "mything" {
      ~ addresses               = [
          - "10.0.1.12/32",
          + "10.0.1.12",
          + "10.10.0.129",
        ]
        id                      = "*1"
        name                    = "mything"
        # (7 unchanged attributes hidden)
    }

```

**Debug Information**

**Stack Trace**

**Additional context**

Thanks so much for all this work. I would like to submit a patch but I'm a newbie to Terraform and provider development, see below.

ROS v7.15: I went to do an SNMP community setup and realized the config with comma-separated ips in cidr notation doesn't work e.g. `"1.1.1.1,8.8.8.8"` or `"1.1.1.1/8,8.8.8.8/32"` doesn't work. It seems like it's accepting a string, which is fine, but because it's using the ip address validator, the comma-separated list fails.

I did manage to set up a basic dev environment and tried to steal some of the `snmp-trap` schema that uses a list of addresses into the `address` part of the `routeros_snmp_community` definition and a full test with hardware _did_ work, but I wasn't able to figure out how to address the `Default` value and didn't see any default values with lists or sets in the project.

Anyway, I'd love to contribute but may need a little help on-boarding.

johnelliott:
Fixes #495

This uses the `Any` ability of the Terraform validation library to do either bare ipv4 or ipv6 addresses (which by observing the device essentially become /32 networks) and CIDR networks.

I wasn't sure what to update test-wise, could use help there.

### DIFF
--- a/docs/resources/snmp_community.md
+++ b/docs/resources/snmp_community.md
@@ -13,6 +13,7 @@ resource "routeros_snmp_community" "test" {
 	name                    = "private"
 	read_access             = true
 	security                = "private"
+	addresses                = ["192.0.2.0/24", "198.51.100.1", "::" ]
 	write_access            = true
 }
 ```
@@ -22,7 +23,7 @@ resource "routeros_snmp_community" "test" {
 
 ### Optional
 
-- `addresses` (String) Addresses from which connections to SNMP server is allowed.
+- `addresses` (Set of String) Set of IP (v4 or v6) addresses or CIDR networks from which connections to SNMP server are allowed.
 - `authentication_password` (String, Sensitive) Password used to authenticate the connection to the server (SNMPv3).
 - `authentication_protocol` (String) The protocol used for authentication (SNMPv3).
 - `comment` (String)
--- a/routeros/resource_snmp_community.go
+++ b/routeros/resource_snmp_community.go
@@ -30,11 +30,17 @@ func ResourceSNMPCommunity() *schema.Resource {
 		MetaId:           PropId(Id),
 
 		"addresses": {
-			Type:         schema.TypeString,
-			Optional:     true,
-			Default:      "::/0",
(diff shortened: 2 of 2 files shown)


johnelliott:
Hi @vaerh this is my first PR. Thanks for the work on this provider, it's really cool!

Is this a direction we could go with this resource? I tested on my little device running v7.15.1.

vaerh:
@johnelliott 
All of this is only moving because of people who care. Thanks for the fixes!
P.S. There is no need to correct the documentation, it is generated automatically at each release.

peterbaumert:
Hi, this introduced an error when you dont specify addresses at all ( since its optional that should be ok )
But it says "Error: missing expected [".

If I add `addresses = []` for example its fine, but i dont know if that will cause other issues.

peterbaumert:
Hi @vaerh actually it seems like an issue of converting existing states with the string to the [] part. I am running ROS 7.15.1

johnelliott:
Thanks @vaerh, I wasn't aware of state migrators until just now and I'll take that into account on my next PR.

PR #502, opened 2024-07-02
feat: Add some new properties introduced in RouterOS 7.15

dokmic:
This PR is mostly about adding newly added properties in the 7.15 release:
- `sfp_ignore_rx_los` to `routeros_interface_ethernet`;
- `enable_ipv6_accounting` to `routeros_ppp_aaa`;
- `require_message_auth` to `routeros_radius`;
- `require_message_auth` to `routeros_user_manager_settings`;
- `reselect_interval` to `routeros_wifi_channel`;
- `current_caps_man_address` to `routeros_wifi_cap`;
- `current_caps_man_identity` to `routeros_wifi_cap`.

On top of that, RouterOS 7.15 started _always_ returning an empty value in `routeros_wifi_configuration` (`/rest/interface/wifi/configuration`):
```
"aaa.password-format": "",
```

Since checking empty old values in the `AlwaysPresentNotUserProvided` helper makes no sense in the first place, we can safely remove this condition. The removal didn't cause any regressions.

### DIFF
--- a/routeros/provider_schema_helpers.go
+++ b/routeros/provider_schema_helpers.go
@@ -671,10 +671,6 @@ var (
 	// Prevents the need of hardcode values for default values, as those are harder to track over time/versions of
 	// routeros
 	AlwaysPresentNotUserProvided = func(k, old, new string, d *schema.ResourceData) bool {
-		if old == "" {
-			return false
-		}
-
 		value := d.GetRawConfig()
 
 		// For lists and sets, the key will look like `something.12345` or `something.#`.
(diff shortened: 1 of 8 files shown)


vaerh:
>Since checking empty old values in the AlwaysPresentNotUserProvided helper makes no sense in the first place, we can safely remove this condition. The removal didn't cause any regressions.

@dokmic does it make sense to remove this check? It allows us to make a fast decision at the first step when checking a large number of new values.

dokmic:
> does it make sense to remove this check? It allows us to make a fast decision at the first step when checking a large number of new values.

@vaerh I don't think keeping it in the first place is correct. The property should be skipped if it's not present in the user configuration, regardless of the stored value. From what I see in the git history, this check was only for such optimization, but it may cause some side effects if we use this helper more.

For RouterOS, it's common to return empty strings for unset properties. This means that empty strings will end up in the state and, hence, in the `old` value. The `aaa.password-format` property is just our first case.

vaerh:
Yeah, I agree.

vaerh (approved):

PR #508, opened 2024-07-17
DRAFT change routeros_wifi_provisioning.radio_mac default to empty string

danielskowronski:
Fix/bypass for bug in 7.16 beta4 (and some earlier releases) on mmips64. 

As per https://help.mikrotik.com/docs/pages/viewpage.action?pageId=1409149#APController(CAPsMAN)-RadioProvisioning the `00:00:00:00:00:00` should be equivalent to null or omitted value, but it's actually setting MAC to that value in config and provisioning fails (random guess - it's actually looking for radio with that MAC). Had to implement in code as both validator and default value made it impossible to bypass inside TF (clearing from CLI works, updating in WebFig magically removes the value) :/

Bug was reported to MikroTik support, but due to nature of TF providers this PR is kind of "FYI bug report". 

This is the only way to use CAPsMAN on RouterBoard that does not have Wi-Fi 6 capabilities (e.g. hEX on mmips64) and control CAPs using Wi-Fi 6 (`wifi-qcom`).

### DIFF
--- a/routeros/resource_wifi_provisioning.go
+++ b/routeros/resource_wifi_provisioning.go
@@ -54,20 +54,20 @@ func ResourceWifiProvisioning() *schema.Resource {
 			Description: "Regular expression to match radios by router identity.",
 		},
 		"master_configuration": {
-			Type:        schema.TypeString,
-			Optional:    true,
+			Type:     schema.TypeString,
+			Optional: true,
 			Description: "If action specifies to create interfaces, then a new master interface with its configuration " +
 				"set to this configuration profile will be created.",
 		},
 		"name_format": {
-			Type:         schema.TypeString,
-			Optional:     true,
-			Description:  "Specify the format of the CAP interface name creation.",
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "Specify the format of the CAP interface name creation.",
 		},
 		"radio_mac": {
 			Type:         schema.TypeString,
 			Optional:     true,
-			Default:      "00:00:00:00:00:00",
+			Default:      "",
 			Description:  "MAC address of radio to be matched, empty MAC (00:00:00:00:00:00) means match all MAC addresses.",
 			ValidateFunc: ValidationMacAddress,
 		},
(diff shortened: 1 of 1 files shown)


vaerh:
Hello!

I apologize for the long absence, I've been sick. I hope to recover soon and get back to the code. Please wait a little while.

routeros/resource_wifi_provisioning.go
```diff
@@ -54,20 +54,20 @@ func ResourceWifiProvisioning() *schema.Resource {
...
+			Description: "Specify the format of the CAP interface name creation.",
 		},
 		"radio_mac": {
 			Type:         schema.TypeString,
 			Optional:     true,
-			Default:      "00:00:00:00:00:00",
+			Default:      "",
 			Description:  "MAC address of radio to be matched, empty MAC (00:00:00:00:00:00) means match all MAC addresses.",
 			ValidateFunc: ValidationMacAddress,
 		},
```
dokmic, line 73: I think the correct approach would be to ignore the value unless provided so that we can make it backward-compatible. No `Default` property in this case.

```suggestion
			Description:  "MAC address of radio to be matched, empty MAC (00:00:00:00:00:00) means match all MAC addresses.",
			ValidateFunc: ValidationMacAddress,
			DiffSuppressFunc: AlwaysPresentNotUserProvided,
		},
```

danielskowronski:
Update from MikroTik support - it's not a bug, it's a feature 🙃 

[WiFi CAPsMAN only treats unset value of `radio-mac` as catch-all](https://help.mikrotik.com/docs/display/ROS/WiFi#WiFi-CAPsMANProvisioning), as opposed to [Wireless CAPsMAN, where "empty MAC (00:00:00:00:00:00) means match all MAC addresses"](https://help.mikrotik.com/docs/pages/viewpage.action?pageId=1409149#APController(CAPsMAN)-RadioProvisioning). 

Therefore, it should be fixed permanently and handle un-setting of value in REST API.

routeros/resource_wifi_provisioning.go
danielskowronski: I agree. I'll test it more today or tomorrow, I've discovered I had some issues when running provider after committing - especially when dealing with entries created with upstream. 

One thing I also noticed was the fact provider seemed to partially behave like WebFig: it didn't treat unsetting `radio-mac` as configuration drift if it was `00:00:00:00:00:00` before. I'll confirm it.

vaerh:
Thanks for your efforts! Not everyone gets the hands to shake a little MT tech support at the strange behavior of this provider :)

@danielskowronski @dokmic 
I don't have CAPsMAN at hand at the moment, so please tell me what we are doing? I would remove the default values and add `DiffSuppressFunc`.

routeros/resource_wifi_provisioning.go
vaerh: @danielskowronski 
Daniel, can I make the changes that Michael suggested?

PR #595, opened 2024-11-12
fix(mangle-connection-state): typo in `established` value

moleus:


vaerh (approved):

vaerh:
Thanks, we'll fix it now...

PR #603, opened 2024-11-21
docs: fix mlag import example

jameslegg:
Import example was wrong for the `routeros_bridge_mlag`.


vaerh (commented):
Thank you very much for the fix!
With your allowance, I will make the example look standard.

PR #632, opened 2025-01-02
feat(tool): add new resource `routeros_tool_email`

heuels:
Hi! Needed this in my setup, figured it would be a nice thing to have in this provider.
I'm not very fluent at Go, so please let me know if anything is wrong. Thanks for your time reviewing this!

Compiled and tested this locally.

### DIFF
--- a/routeros/resource_tool_email.go
+++ b/routeros/resource_tool_email.go
@@ -24,10 +24,10 @@ func ResourceToolEmail() *schema.Resource {
 		MetaId:           PropId(Id),
 
 		"from": {
-			Type:        schema.TypeString,
-			Optional:    true,
-			Description: "Name or email address that will be shown as a receiver.",
-			Default:     "<>",
+			Type:             schema.TypeString,
+			Optional:         true,
+			Description:      "Name or email address that will be shown as a receiver.",
+			DiffSuppressFunc: AlwaysPresentNotUserProvided,
 		},
 		"password": {
 			Type:        schema.TypeString,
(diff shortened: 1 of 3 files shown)


routeros/resource_tool_email.go
```diff
@@ -0,0 +1,79 @@
...
+func ResourceToolEmail() *schema.Resource {
+	resSchema := map[string]*schema.Schema{
+		MetaResourcePath: PropResourcePath("/tool/e-mail"),
+		MetaId:           PropId(Id),
+
+		"from": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "Name or email address that will be shown as a receiver.",
+			Default:     "<>",
```
vaerh, line 30: Thank you very much for your contribution!

I have a request to replace all default values with `DiffSuppressFunc: AlwaysPresentNotUserProvided,`, we have thus solved the problem of tracking default values between different ROS versions.
This should also be done for: `port`, ``server`, `tls`.

vaerh (changes requested):

routeros/resource_tool_email.go
heuels: Appreciate the review! I just committed these changes.

heuels:
I also created resources today for SSTP server and client, will test and clean them up a bit tomorrow and open a PR 🙏🏻

vaerh (approved):

vaerh (approved):

PR #642, opened 2025-01-12
fix(datasource_routeros_interface_bridge_filter): Returned data key should be `filters`

DionJones615:
Fixes a bug that results in an empty data object every time due to: `[MikrotikResourceDataToTerraformDatasource] the datasource Schema field was lost during development: ▷ 'data' ◁`

[Schema](https://github.com/terraform-routeros/terraform-provider-routeros/blob/a97fe34287072f1993e8134ab31f3909d37fe024/routeros/datasource_interface_bridge_filter.go#L20) shows the key is `filters`


vaerh (approved):
An unfortunate typo, thanks so much for the fix!

Issue #652, opened 2025-02-04
Importing containers and container mounts (inconsistency and force recreate)


MagicMicky:
When importing containers and containers mounts I noticed two issue, one being an inconsistency or documentation issue, the second being a potential inability to import

Container mounts:
[As per documentation](https://registry.terraform.io/providers/terraform-routeros/routeros/latest/docs/resources/container_mounts) and in consistency with other resources, importing should be done with mikrotik internal resource ID - however it's actually expected to be done with the mount name https://github.com/terraform-routeros/terraform-provider-routeros/blob/main/routeros/resource_container_mounts.go#L11

For  containers, importing works as expected, except it wants to recreate the resource from scratch. From what I seem to understand it is because remote_image wants to create a new container. It seems to move to "tag" after being created.

Example:
* Create container through web interface
* Get it's id as per documentation
* Import  `terraform import -var-file creds.tfvars routeros_container.agh_container "*F"`
* Plan

```Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # routeros_container.agh_container must be replaced
-/+ resource "routeros_container" "agh_container" {
      ~ arch          = "arm64" -> (known after apply)
      ~ id            = "*F" -> (known after apply)
      ~ name          = "a9e34742-c27e-42eb-a9ab-f274364d18ed" -> (known after apply)
      ~ os            = "linux" -> (known after apply)
      + remote_image  = "adguard/adguardhome:latest" # forces replacement
      + start_on_boot = false
      ~ status        = "running" -> (known after apply)
      ~ tag           = "adguard/adguardhome:latest" -> (known after apply)
      ~ workdir       = "/opt/adguardhome/work" -> (known after apply)
        # (4 unchanged attributes hidden)
    }
```

Thanks for all the work on this provider :)

vaerh:
I will look into the problem in the near future. I can't say anything yet as the code is not mine.

vaerh:
@MagicMicky 
Hi Mickael!

I finally deployed the bench at CHR with containers, I apologize for the delay. It was a bit of a problem.
The debugging results are not very good yet. MT doesn't return the `remote_image` parameter and for this reason the provider doesn't work properly.
I will try to write to the developers on the forum and see what happens.

vaerh:
@MagicMicky 
Hi!

[...]

PR #677, opened 2025-03-01
WIP: allow importing containers without unnecessary recreating

ksmets:
As mentioned in #652, there is an inconsistency in the documenation on how to properly import container mounts.

Moreover, a suggestion implementation on how to use computed fields `tag` (ROS < 7.18), `repo` (ROS >=7.18) to repopulate the `remote_image` write-only field.

### DIFF
--- a/examples/resources/routeros_container_mounts/import.sh
+++ b/examples/resources/routeros_container_mounts/import.sh
@@ -1,3 +1,2 @@
-#The ID can be found via API or the terminal
-#The command for the terminal is -> :put [/container/mounts get [print show-ids]]
-terraform import routeros_container_mounts.caddyfile "*1"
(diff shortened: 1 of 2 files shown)


vaerh:
I have a big request to wait a bit, I won't be at my laptop today.

vaerh:
@ksmets 
Hi!
It turned out to be much more complicated than it was at first glance. At the moment we cannot use `WriteOnly` fields together with `ForceNew`. Because of this, the provider stops responding to changes in the `remote_image` attribute. I had to write some code and I have a request to check if everything works well.
#680

P.S. I added the possibility of attribute drift between ROS versions. To do this, you can now just fill in the `mikrotik_resource_drift.yaml` file without changing the schema.

PR #729, opened 2025-05-01
Implement duration check

Elbandi:
Fix wireguard peer persistent-keepalive validation: it's time duration (5s, 10m, 1h, etc).

You can set a decimal value, but mikrotik will convert it to duration, so terraform always want to change it to integer back.


routeros/resource_interface_wireguard_peer.go
```diff
@@ -102,15 +99,14 @@ func ResourceInterfaceWireguardPeer() *schema.Resource {
 		"persistent_keepalive": {
 			Type:     schema.TypeString,
 			Optional: true,
+			Default:  "0",
```
vaerh, line 102: Thank you very much for your fix! 
The only request is to replace the `Default` value with `DiffSuppressFunc: AlwaysPresentNotUserProvided,`.

vaerh (changes requested):

routeros/resource_interface_wireguard_peer.go
Elbandi: replaced in new commit

vaerh (approved):

PR #758, opened 2025-06-01
fix: bgp connection property address-family has been renamed to afi

Tchoupinax:
Hey folks,

It seems that in `resource_bgp_connection` object a property has been renamed and introduced a breaking change.
This pull request aims to fix this property.

I did not find anything in changelog but I identified [this change in the documentation ](https://help.mikrotik.com/docs/pages/diffpagesbyversion.action?pageId=328220&selectedPageVersions=79&selectedPageVersions=78) that shows the rename of the property. Also, I compiled these modifications locally and applied my code, it works well (and so fix the issue I had).

**Current Behavior**

API rejects requests with 400 because `address_families` is unknown

**Expected Behavior**

API accepts requests and creates resources

Thanks!

### DIFF
--- a/docs/resources/routing_bgp_connection.md
+++ b/docs/resources/routing_bgp_connection.md
@@ -14,7 +14,7 @@
 ### Optional
 
 - `add_path_out` (String)
-- `address_families` (String) List of address families about which this peer will exchange routing information. The remote peer must support (they usually do) BGP capabilities optional parameter to negotiate any other families than IP.
+- `afi` (String) List of address families about which this peer will exchange routing information. The remote peer must support (they usually do) BGP capabilities optional parameter to negotiate any other families than IP.
 - `cisco_vpls_nlri_len_fmt` (String) VPLS NLRI length format type. Used for compatibility with Cisco VPLS.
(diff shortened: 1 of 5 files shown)


Tchoupinax:
Hey @vaerh,
Thanks for the tip!

I reverted changes and added the drift inside the file but I probably need assistance on that because I tested the provider and I can't achieve the same result (as I did with changed code).

I saw that BGP routing seems to use defaultUpdate so I guess drift is applied. What do I miss?

vaerh:
Oh, sorry!

I forgot to warn you that it is necessary to start generation of a file with drift parameters:

``` bash
cd routeros/
go run ../tools/drift/main.go
```

By the way, this attribute has changed since version 7.19:

``` yaml
7.19:
  /routing/bgp/connection:
    - tf: address_families
      mt: afi
  /routing/bgp/template:
    - tf: address_families
      mt: afi
```

Can you please check if everything will work correctly with these settings?

vaerh:
Everything is perfect, checking it out and getting the release ready!

vaerh (approved):

vaerh:
You're welcome!

PR #785, opened 2025-07-25
feat: Implement RouterOS TFTP resources

### LINKED ISSUE #784: Resource for Managing TFTP missing

I want to setup a [PXE](https://en.wikipedia.org/wiki/Preboot_Execution_Environment) with my router and [matchbox](https://matchbox.psdn.io/). This requies to have a DHCP server, which supportes TFTP to allow the PXE clients to download the boot images.

I can already upload the boot images to the Router using the `routeros_file` resources. What I'm mssing now is are the resource definitions for the terraform provider, to define the TFTP files the clients are allowed to download.

According to the naming scheme, this would be something like `routeros_ip_tftp`.

**Documentation**
 RouterOS already has support for it built-in under `/ip/tftp` (see https://help.mikrotik.com/docs/spaces/ROS/pages/131366922/TFTP).

hansingt:
Closes #784.

I've added two new resources to the provider: `routeros_ip_tftp` and `routeros_ip_tftp_settings`.

The former can be used to add entries to the TFTP server provided by RouterOS.

The later one can be used to configure the TFTP server. **For this, only a single resource can be defined as otherwise, they will overwrite each other!**


vaerh:
Thank you very much for your work!

vaerh:
I have no doubts about the performance of your code, in no case!
With your permission, I will bring it to the form of other resources.

vaerh (approved):

PR #794, opened 2025-07-28
Add DefaultReadFiltered, filter `ip_dhcp_server_lease` resource reads based on `dynamic` field

kahlstrm:
The setup that I have currently is that I have a bootstrap script that sets up the devices in a state where I can run `terraform apply` against that, regardless of the existing state in terraform. This is admittedly a bit cursed but bear with me.

The problem I'm facing is with dhcp leases, where MikroTik seemingly doesn't separate between dynamic leases and static ones in their IDs.
This then causes a problem where the ID that terraform state has for the static lease is currently a dynamic lease, which results in `Error: PATCH 'https://10.1.1.2/rest/ip/dhcp-server/lease/*1' returned response code: 400, message: 'Bad Request', details: 'failure: can not change dynamic lease'`.

As dynamic leases can not be changed anyways, probably would be okay to add filtering on the read of `routeros_ip_dhcp_server_lease` that filters out dynamic leases with `dynamic=no`.

Did this with adding a `DefaultReadFiltered` as well as two more underlying functions to mimic `DefaultRead`, and then calling it `routeros/resource_ip_dhcp_server_lease.go`.

Tested this at least locally and it works for my use case and doesn't seem to cause any weirdness.

Please let me know if this makes any sense, I'm new to terraform providers so there's also that.

### DIFF
--- a/routeros/resource_actions.go
+++ b/routeros/resource_actions.go
@@ -255,6 +255,35 @@ func ResourceRead(ctx context.Context, s map[string]*schema.Schema, d *schema.Re
 	return MikrotikResourceDataToTerraform((*res)[0], s, d)
 }
 
+// ResourceReadFilter Reading some information about one specific resource with additional filters.
+func ResourceReadFilter(ctx context.Context, s map[string]*schema.Schema, d *schema.ResourceData, m interface{}, filters map[string]interface{}) diag.Diagnostics {
+	metadata := GetMetadata(s)
+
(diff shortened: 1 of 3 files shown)


vaerh:
Not bad:)
Please tell me what is the use case when static lease is not static?
Are you importing existing resources?

vaerh:
This question is for Terraform ;)
I think everything else should work.

PR #853, opened 2025-10-30
feat: (routeros_interface_ethernet_switch) - Add missing attribute

rpchan44:


routeros/resource_interface_ethernet_switch.go
```diff
@@ -115,6 +115,50 @@ func ResourceInterfaceEthernetSwitch() *schema.Resource {
 		// 		"88E6190 switch chips. Mirrored packets from mirror-egress (see the property in port menu) will be " +
 		// 		"sent to the selected port.",
 		// },
+		"not_learned": {
```
vaerh: Ok, very good for the first iteration! But you need to make small changes :)

For the convenience of searching, the attributes in most cases are arranged in alphabetical order, since with their large number, the scheme should not turn into chaos.

https://mikrotik.wiki/wiki/MikroTik_RouterOS_6.48rc1_(Testing)
`crs3xx - fixed "custom-drop-packet" and "not-learned" switch stats for CRS317, CRS309, CRS312, CRS326-24S+2Q+ and CRS354 devices;`
Judging by the description, this field and the following are statistics, respectively, in the provider there are two cases of specifying such attributes:
* if the returned data may be needed in a Terraform (TF) script, then we create an attribute and specify `Computed: true,` as in `type`
* if the returned data is useless, then it is better to enter it in `PropSkipFields`.
For these fields, I would take the second path.

routeros/resource_interface_ethernet_switch.go
```diff
@@ -115,6 +115,50 @@ func ResourceInterfaceEthernetSwitch() *schema.Resource {
...
+			Type:     schema.TypeBool,
+			Optional: true,
+			Default:     false,
+			Description: "Custom drop packet",
+			DiffSuppressFunc: AlwaysPresentNotUserProvided,
+		},
+		"qos_hw_offloading": {
+			Type:     schema.TypeBool,
+			Optional: true,
+			Default:     false,
```
vaerh: To avoid problems with different versions of ROS, we use ёAlwaysPresentNotUserProvidedё instead of `Default` value. Moreover, if the attribute returns `false`, `0`, empty value or absent in the MikroTik (MT) response, then the `AlwaysPresentNotUserProvided` function may not be specified.
In this case, `Default` and `AlwaysPresentNotUserProvided` can be removed.

vaerh: [Description](https://help.mikrotik.com/docs/spaces/ROS/pages/189497483/Quality+of+Service)
qos-hw-offloading (yes | no; Default: no)	Allows enabling QoS for the given switch chip (if the latter supports QoS).

routeros/resource_interface_ethernet_switch.go
```diff
@@ -115,6 +115,50 @@ func ResourceInterfaceEthernetSwitch() *schema.Resource {
...
+			DiffSuppressFunc: AlwaysPresentNotUserProvided,
+		},
+		"rspan": {
+			Type:     schema.TypeBool,
+			Optional: false,
+			Default:     false,
+			Description: "RSPAN Toggle",
+			DiffSuppressFunc: AlwaysPresentNotUserProvided,
+		},
+		"rspan_ingress_vlan_id": {
```
vaerh: In this attribute and the following, you need to delete `Default`
Description in the previous link.

vaerh:
In order not to bother you, I myself made small corrections.

PR #865, opened 2025-11-15
feat: routeros_interface_macvlan add missing attributes

rpchan44:


rpchan44:
Why it's failing the test?

PR #897, opened 2025-12-12
Move dynamic IP/services to new attr on data source

### LINKED ISSUE #896: data routeros_ip_services needs to be able to filter out connections/dynamic services

RouterOS 7.19 introduced current connections and other 'dynamic' services (such as DNS server) to the IP services list. These are characterised by `"connection": "true"` and `"dynamic": "true"` (strings sic) on the REST API, though connection seems simply absent when false.

This breaks the documented example of this data source:
```tf
data "routeros_ip_services" "router" {
  provider = routeros.router
}

resource "routeros_ip_service" "router-disabled" {
  provider = routeros.router
  for_each = { for s in data.routeros_ip_services.router.services : s.name => s if s.name != "www-ssl" }

  disabled = true
  numbers  = each.value.name
  port     = each.value.port
}
```

since there will likely be multiple occurrences of `s.name = "resolver"` present, and I suppose certainly multiple `s.name = "www"` or as applicable to whatever terraform is configured to use for the API.

> │ Error: Duplicate object key
> │
> │   on main.tf line 139, in resource "routeros_ip_service" "router-disabled":
> │  139:   for_each = { for s in data.routeros_ip_services.router.services : s.name => s if s.name != "www-ssl" }
> │     ├────────────────
> │     │ s.name is "resolver"
> │
> │ Two different items produced the key "resolver" in this 'for' expression. If duplicates are expected, use the ellipsis
> │ (...) after the value expression to enable grouping by key.

`data.routeros_ip_services` does support `filter`, but `connection` & `dynamic` are not presently deserialised and so can't be filtered.

`connection` and `dynamic` could be deserialised and filterable, but I actually don't think there's any reasonable and useful use for those 'services'; I would argue that `connection`s _at least_ should be automatically excluded, perhaps a new data source for them (`routeros_ip_services_dynamic`) or read-only attribute (`router_os_ip_services.router.connections`) if there is a use case.

**Documentation**
It doesn't seem to be documented, just a brief changelog note: https://forum.mikrotik.com/t/v7-19-6-stable-is-released/183849

> * ip-service - show all TCP/UDP connections on the system;
> * ip-service - show all TCP/UDP ports on system, including ports in containers;

**Additional context**

OJFord:
This fixes a regression in `data.routeros_ip_services` in that 7.19 introduced 'dynamic' services & connections there, and `name` is no longer (with those included) unique. They're also much less if at all useful from a configuration perspective.

Here, I've (1) excluded them from `data.routeros_ip_services.*.services`, so that it behaves as it did pre-7.19; (2) added `data.routeros_ip_services.*.dynamic_services`, with exactly those excluded from the former; (3) added `proto` (TCP/UDP) and `max_sessions` fields to the schema of both, which was causing warnings otherwise. (I'm not sure which ROS version introduced those.)

Closes #896.

### DIFF
--- a/docs/data-sources/ip_services.md
+++ b/docs/data-sources/ip_services.md
@@ -64,8 +64,10 @@ Read-Only:
 - `dynamic` (Boolean)
 - `id` (String)
 - `invalid` (String)
+- `max_sessions` (Number)
 - `name` (String)
 - `port` (Number)
+- `proto` (String)
 - `tls_version` (String)
 - `vrf` (String)
 
--- a/routeros/datasource_ip_services.go
+++ b/routeros/datasource_ip_services.go
@@ -116,12 +116,20 @@ func DatasourceIPServices() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
+						"max_sessions": {
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
 						"name": {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
 						"port": {
 							Type:     schema.TypeInt,
+							Computed: true,
+						},
+						"proto": {
+							Type:     schema.TypeString,
 							Computed: true,
 						},
 						"tls_version": {


vaerh (approved):

OJFord:
Thanks!

PR #901, opened 2025-12-17
Fix ambiguous (null) stop signal

OJFord:
When unspecified, the provider would try to remove the ROS default of `"15-SIGTERM"` on second apply.

### DIFF
--- a/routeros/resource_container.go
+++ b/routeros/resource_container.go
@@ -174,6 +174,7 @@ func ResourceContainer() *schema.Resource {
 		"stop_signal": {
 			Type:        schema.TypeString,
 			Optional:    true,
+			Default:     "15-SIGTERM",
 			Description: "Signal to stop the container.",
 		},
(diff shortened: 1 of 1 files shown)


routeros/resource_container.go
```diff
@@ -174,6 +174,7 @@ func ResourceContainer() *schema.Resource {
 		"stop_signal": {
 			Type:        schema.TypeString,
 			Optional:    true,
+			Default:     "15-SIGTERM",
```
vaerh: Instead of the `Default` value, we add `DiffSuppressFunc: AlwaysPresentNotUserProvided,`

vaerh (changes requested):

routeros/resource_container.go
vaerh: Ok, no problem. Then let's add it as is.

vaerh (approved):

PR #910, opened 2026-01-01
Fix container mounts and misc. new attrs in ROS v7.21

OJFord:
Mounts are now in mount lists, though individual mounts can additionally
be specified directly on the container definition (not implemented
here).

Env vars (already in lists) can also now be specified directly on the container, but I haven't implemented that either.

Aside from those, I've added all the new or renamed attrs that I saw from a plan & simple update; there's no doubt more though, I just haven't run into them yet.


docs/resources/container.md
```diff
@@ -26,30 +26,40 @@ resource "routeros_container" "busybox" {
...
+- `env` (Set of String) list of environmental variables (in the form `key=val`) to be used with container
+- `envlists` (Set of String) list of environmental variables lists (configured under /container envs) to be used with container
 - `file` (String) container *tar.gz tarball if the container is imported from a file
 - `hostname` (String) Container host name
+- `hosts` (String)
+- `layer_dir` (String) Override container config layer dir
 - `logging` (Boolean) if set to yes, all container-generated output will be shown in the RouterOS log
 - `memory_high` (String) RAM usage limit in bytes for a specific container (string value).
-- `mounts` (Set of String) Mounts from /container/mounts/ sub-menu to be used with this container
+- `mount` (Set of String) Mounts (in the form `/src:/mnt:rw`) to be used with this container
```
elbandi: This is breaks backward compatible?

routeros/resource_system_certificate.go
```diff
@@ -366,11 +366,17 @@ func ResourceSystemCertificate() *schema.Resource {
 			Description:      "SANs (subject alternative names).",
 			DiffSuppressFunc: AlwaysPresentNotUserProvided,
 		},
+		"trust_store": {
```
vaerh: Doesn't work on older versions of ROS.

routeros/resource_container_mounts.go
```diff
@@ -8,12 +8,29 @@ import (
 func ResourceContainerMounts() *schema.Resource {
 	resSchema := map[string]*schema.Schema{
 		MetaResourcePath: PropResourcePath("/container/mounts"),
-		MetaId:           PropId(Name),
+		MetaId:           PropId(Id),
```
vaerh: This is a very bad change. If I apply it, the old resources will be lost and will need to be imported. Otherwise, I'll have to update the schema. I need to check and think about it...

docs/resources/container_mounts.md
```diff
@@ -16,15 +16,17 @@ resource "routeros_container_mounts" "caddyfile" {
 ### Required
 
 - `dst` (String) Specifies destination path of the mount, which points to defined location in container
-- `name` (String) Name of the mount.
+- `list` (String) Mount list this mount is for
```
elbandi, line 19: example is not synced (name -> list)

PR #915, opened 2026-01-11
Fix some erroneously suppressed diffs with proper defaults

OJFord:
I appreciate the intent of `AlwaysPresentNotUserProvided`, but I think in at least most cases it unfortunately can't be used, and is causing erroneous diff suppressions where it is:

Suppose you apply a filter rule with `log=true`, check it out in WebFig, decide it's good, and untick `log` in the UI. On next `terraform apply` you *should* see it want to:
```
  log = false -> true
```
but due to `AlwaysPresentNotUserProvided`, prior to this commit, it will suppress that diff and not want to revert the manual change.

Similarly, if `mdns_repeat_ifaces` is not initially specified, then added in the UI, terraform will continue ignoring it until/unless explicitly added in terraform config.

In my view this makes for quite a gotcha: anything can happen to parameters that are not explicitly specified in configuration, the plan will look clean, and yet they're not in code at all, but may well be relied upon values.

### DIFF
--- a/routeros/resource_interface_bridge.go
+++ b/routeros/resource_interface_bridge.go
@@ -104,12 +104,11 @@ func ResourceInterfaceBridge() *schema.Resource {
 		"ingress_filtering": {
 			Type:     schema.TypeBool,
 			Optional: true,
+			Default:  true,
 			Description: "Enables or disables VLAN ingress filtering, which checks if the ingress port is a member " +
 				"of the received VLAN ID in the bridge VLAN table. Should be used with frame-types to specify if " +
 				"the ingress traffic should be tagged or untagged. This property only has effect when " +
 				"vlan-filtering is set to yes.",
-			DiffSuppressFunc: AlwaysPresentNotUserProvided,
-			RequiredWith:     []string{"vlan_filtering"},
 		},
 		KeyL2Mtu: PropL2MtuRo,
 		"last_member_interval": {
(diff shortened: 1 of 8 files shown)


vaerh:
My business is to dissuade everyone from managing resources from outside:) There are architectural problems with MT.

I don't mind living with such changes, but then for complete uniformity we need to include the default value for the entire scheme. Otherwise, we risk getting more inconsistency when working with the resource.

OJFord:
I completely agree everything should be in IaC and in a single tool, i.e. only terraform if using terraform.

But they should at least be `Computed: true`, so that the remote value is retained in tf state, and can be referenced appropriately (with value 'known after apply', not `null`) in other resources.

I think in an ideal world, one where MT published an API schema or go package(!), we would use `DefaultFunc` to provide a default based on the ROS version in use. In this world though, maybe we could just switch on a provider-level attribute `version = "stable" | "lts"`, and via something similar to the drift YAML spec attempt to track LTS version differences from the rolling changes for stable, but not have to do so for _every_ version of ROS. What do you think?

routeros/resource_interface_bridge.go
```diff
@@ -104,12 +104,11 @@ func ResourceInterfaceBridge() *schema.Resource {
 		"ingress_filtering": {
 			Type:     schema.TypeBool,
 			Optional: true,
+			Default:  true,
 			Description: "Enables or disables VLAN ingress filtering, which checks if the ingress port is a member " +
 				"of the received VLAN ID in the bridge VLAN table. Should be used with frame-types to specify if " +
 				"the ingress traffic should be tagged or untagged. This property only has effect when " +
 				"vlan-filtering is set to yes.",
-			DiffSuppressFunc: AlwaysPresentNotUserProvided,
-			RequiredWith:     []string{"vlan_filtering"},
```
vaerh: Once again, I ask for forgiveness for my departure beyond the planet Earth)))
I had a small window and a little strength and I wanted to merge some requests.
In this regard, I have a question, was there any reason to delete'`RequiredWith: [] string {"vlan _ filtering"}` ?

PR #935, opened 2026-02-19
Add attributes to ipv6_address and ip_ssh_server [7.21]

parcimonic:
For IPv6 address `vrf` I followed the same approach as: https://github.com/terraform-routeros/terraform-provider-routeros/blob/a1d75fd8dfbe3799c06ae5a35a575acbfaa5b4fb/routeros/resource_ip_address.go#L57

I imagine it is related to this change in 7.21 changelog:

> *) vrf - added read-only property to IPv4/IPv6 addresses, ARP and IPv6 neighbor;

Should it be also added to the data sources though?

---

For SSH server `publickey_authentication_options`, I couldn't find it in the changelog but in their Confluence docs it was updated last month: https://help.mikrotik.com/docs/pages/diffpagesbyversion.action?pageId=132350014&selectedPageVersions=23&selectedPageVersions=22

---

I've tested the `tfplugindocs` tool to update the documentation, hopefully I've got it right.

### DIFF
--- a/routeros/resource_ip_ssh_server.go
+++ b/routeros/resource_ip_ssh_server.go
@@ -68,13 +68,13 @@ func ResourceIpSSHServer() *schema.Resource {
 		"publickey_authentication_options": {
 			Type:     schema.TypeString,
 			Optional: true,
-			Default:  "none",
 			Description: "Sets public key authentication options." +
 				"\nThe touch-required option causes public key authentication using a FIDO authenticator " +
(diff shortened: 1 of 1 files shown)


vaerh:
Unfortunately, since the provider doesn't support schema versioning, I had to change the attribute parameters to save backward compatibility. However, this means we lose the ability to respond to attribute changes from outside of TF.

PR #957, opened 2026-03-16
feat: update RouterOS versions in module testing workflow

danielnachtrub:


vaerh:
I would like to ask you not to remove deprecated attributes unless they cause errors during testing. Users are not always updated to the latest version of ROS.

danielnachtrub:
It would be ideal to resolve this version dependency once because the API is really unstable and many failures arise during apply which could be avoided (maybe).

Having the mapping already in the provider _could_ be an option but makes the maintenance much harder. Should we omit those attributes them in the tests or try to make the tests aware of the ROS version?

vaerh:
I'm on the road now, and today I'll write about what I've figured out with versioning. It's not all that simple, but there are some ideas that need to be discussed.

vaerh:
[Let's go](https://github.com/orgs/terraform-routeros/discussions/958)

PR #977, opened 2026-04-19
feat: opt-in bulk_read_refresh for order-of-magnitude faster refresh

### LINKED ISSUE #976: `terraform plan` scales linearly with resource count on a single path — RouterOS REST API serializes requests
# `terraform plan` scales linearly with resource count on a single path — RouterOS REST API serializes requests

## What happens

When a configuration manages many resources on the same RouterOS path with `for_each` or `count` (e.g. many `routeros_dns_record`s, `routeros_ip_firewall_filter` rules, `routeros_ip_firewall_addr_list` entries, `routeros_ip_route`s), `terraform plan` and `terraform refresh` take time proportional to the resource count. The provider issues one filtered `GET /rest/<path>?.id=*X` per resource during refresh, and the MikroTik www service processes those requests **serially** regardless of `-parallelism`, HTTP keep-alive, or transport tuning.

1868 `routeros_dns_record` entries on `/ip/dns/static` refresh in 3m 42s. A single `GET /rest/ip/dns/static` returns the same 1880 records in 1.2s.

## Reproduction

```hcl
resource "routeros_dns_record" "spam" {
  for_each = toset([for i in range(500) : "host${i}.example"])
  name     = each.key
  address  = "127.0.0.1"
  type     = "A"
}
```

`terraform plan` on an already-applied state takes proportionally longer as the range grows.

## Environment

- Provider `terraform-routeros/routeros` v1.99.1
- OpenTofu 1.8.x / Terraform 1.8.x
- MikroTik hEX (RB750Gr3), RouterOS 7.22.1
- Transport: REST over HTTP, default `/ip/service/www max-sessions=20`

## Root cause

Instrumented the provider with `net/http/httptrace`, aggregated over 1800 requests during one plan:

```
SendReq avg=1.355s | httptrace: getConn=17µs write=21µs srvThink(TTFB)=1.355s bodyRead=149µs | reused=1800 newConn=10
```

The connection pool is reused as expected and client-side work adds up to ~200µs per request. The 1.355s is time-to-first-byte — the router queueing the concurrent requests.

Arithmetic:
- 256s total / 1868 requests = 137 ms server-side work per request
- With 10 workers in flight, expected TTFB = 9×137 + 137 = 1.37s ✓

```mermaid
sequenceDiagram
    participant TC as Terraform Core
    participant P as routeros provider
    participant R as RouterOS www

    par Worker 1
        TC->>P: Read(id=*1)
        P->>R: GET /rest/ip/dns/static?.id=*1
    and Worker 2..10
        TC->>P: Read(id=*2..*A)
        P->>R: GET /rest/ip/dns/static?.id=*2..*A
    end

    Note over R: 1 request at a time
    R-->>P: *1 (137ms)
    R-->>P: *2 (274ms)
    R-->>P: *A (1370ms)

    Note over TC,R: × 1868
```

## Prior reports

The RouterOS REST serialization is a known but undocumented quirk:

- [forum.mikrotik.com #266665 — "RouterOS rest API requests running in series"](https://forum.mikrotik.com/t/routeros-rest-api-requests-running-in-series/266665) — parallel REST requests time out; sequential succeeds. RouterOS 7.20.4.
- [forum.mikrotik.com #173735 — "Rest API Limitations (if any)"](https://forum.mikrotik.com/t/rest-api-limitations-if-any/173735) — parallel REST requests crashed the www service on hAP ac3.
- [Official REST API docs](https://help.mikrotik.com/docs/spaces/ROS/pages/47579162/REST+API) — silent on concurrency.
- [Official binary API docs](https://help.mikrotik.com/docs/spaces/ROS/pages/47579160/API) — documents concurrency via the `tag` parameter, but that's port 8728, not REST.

Behavior persists across at least RouterOS 7.20 → 7.22 with no public MikroTik fix.

## Workarounds that don't help

| Config | Wall |
|---|---|
| `-parallelism=1` | 5m 7s |
| `-parallelism=10` | 4m 16s |
| `-parallelism=50` | 4m 18s |
| HTTP `MaxIdleConnsPerHost=1`..`10` | ≈4m 15s (no change) |

Above ~2 concurrent requests, raising parallelism stops improving wall time because the router processes them serially.

## Partial workaround users can apply today

Switching `hosturl` from `http[s]://` to `api://` / `apis://` uses the native binary protocol and saves ~40% (4m 16s → 2m 41s on the same dataset). Still linear in resource count.

## Fix

Issue one bulk `GET <path>` per path and serve subsequent per-id reads from an in-memory cache invalidated on writes. Implemented in #977 behind the opt-in provider attribute `bulk_read_refresh` (default `false`).

Measured with the fix:

| Config | Wall |
|---|---|
| v1.99.1 unmodified | 3m 42s |
| `bulk_read_refresh=true` | **5.7s** (46× faster) |

A second plan after applying the fix reports no changes — refresh produces the same state as the unmodified provider.

The fix sits in the shared `ReadItems` helper, so it applies to every resource that uses `DefaultRead` (~150 resources). Datasource reads and `IdType==Name` lookups bypass the cache, because their semantics differ from the per-id reads the cache is built for.

goto1134:
Fixes #976.

## Summary

State refresh on paths with many resources currently takes time proportional to the resource count because the provider issues one filtered `GET` per resource, and RouterOS's www service processes these serially. See #976 for the full problem analysis, measurements, and prior art.

This PR adds an opt-in provider attribute `bulk_read_refresh` (default `false`, env `ROS_BULK_READ_REFRESH`). When enabled, the first `ReadItems` call on a path triggers one bulk `GET <path>` and caches the response indexed by `.id`. Subsequent per-id reads on the same path serve from memory. Successful `Create`/`Update`/`Delete` on a path invalidates its cache entry.

## Measured impact

Against a live MikroTik hEX with 1868 `routeros_dns_record` entries on `/ip/dns/static`:

| Config | Wall |
|---|---|
| v1.99.1 unmodified | 3m 42s |
| `bulk_read_refresh=true` | 5.7s (~46× faster) |

A second plan reports no changes — refresh produces identical state.

## Design

- Cache keyed by path, indexed by `.id`. Scoped to one provider instance (one terraform command).
- First miss populates via bulk GET; concurrent misses on the same path coalesced by `golang.org/x/sync/singleflight` (already transitive via `terraform-plugin-sdk`).
- Successful `Create`/`Update`/`Delete` invalidate that path's entry. Failed writes do not invalidate (router state is unchanged, so the cache is still correct).
- **Datasource reads** (`id == nil`) and **Name-type lookups** (`IdType == Name`) bypass the cache — their semantics differ from the per-id reads the cache is built for.
- **Transport-neutral**: benefits both REST (`http[s]://`) and native API (`api[s]://`). The fix sits in the shared `ReadItems`/`CreateItem`/`UpdateItem`/`DeleteItem` helpers in `routeros/mikrotik_crud.go`, so every resource using `DefaultRead` inherits it (~150 resources).

## Test plan

- [x] 6 unit tests for `BulkReadCache`: Lookup (hit / negative-hit-on-known-path / empty-cache), Invalidate, singleflight coalescing (50 concurrent callers → 1 bulk fetch), error propagation
- [x] 11 unit tests for `ReadItems`/`CreateItem`/`DeleteItem` using a fake `Client`: cache-disabled → filtered GET, first-call → bulk GET, second-call → 0 requests, id-not-in-bulk → empty, post-invalidate refetch, nil-id bypass, Name-type bypass, bulk-GET error propagation, 100-goroutine coalescing, Create-success invalidation, Delete path-mutation regression guard
- [x] 2 acceptance tests (run with `TF_ACC=1` + `ROS_*` envs) across both REST and API transports: roundtrip (apply + plan-only, expect no drift) and out-of-band drift detection (external DELETE → plan detects drift)
- [x] All 17 unit tests pass 3× under `go test -count=3`
- [x] Verified against real router (1868 records): 3m 42s → 5.7s

## Questions for reviewers

1. Opt-in default, or on-by-default with opt-out? I chose opt-in to preserve existing behavior during rollout.
2. Any resources with `IdType==Name` Read semantics that the cache bypass might still subtly affect? I audited the codebase and found only `ResourceCreate` post-create name-search and `ResourceCreateAndWait` polling — both bypass the cache explicitly.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -62,6 +62,7 @@ resource "routeros_interface_gre" "gre_hq" {
 
 ### Optional
 
+- `bulk_read_refresh` (Boolean) When true, Terraform state refresh fetches each resource path in a single bulk GET and serves subsequent per-id reads from an in-memory cache invalidated on writes. Speeds up plans over large resource sets (e.g. thousands of DNS records) by an order of magnitude when most records on a path are managed by Terraform. Can slow down plans on configs that manage only a handful of records on a path populated by many unmanaged entries, because each bulk fetch returns the full path. Opt-in; default false preserves the traditional per-id filtered GET behavior (env: ROS_BULK_READ_REFRESH).
 - `ca_certificate` (String) Path to MikroTik's certificate authority file (env: ROS_CA_CERTIFICATE | MIKROTIK_CA_CERTIFICATE).
 - `insecure` (Boolean) Whether to verify the SSL certificate or not (env: ROS_INSECURE | MIKROTIK_INSECURE).
 - `password` (String, Sensitive) Password for the MikroTik user (env: ROS_PASSWORD | MIKROTIK_PASSWORD).
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.40.0
 	github.com/sirupsen/logrus v1.9.4
 	github.com/sourcegraph/go-diff-patch v0.0.0-20240223163233-798fd1e94a8e
+	golang.org/x/sync v0.20.0
 )
 
 require (
@@ -30,7 +31,6 @@ require (
 	github.com/yuin/goldmark-meta v1.1.0 // indirect
 	go.abhg.dev/goldmark/frontmatter v0.2.0 // indirect
 	golang.org/x/exp v0.0.0-20230809150735-7b3493d9a819 // indirect
-	golang.org/x/sync v0.20.0 // indirect
 	golang.org/x/term v0.41.0 // indirect
 	golang.org/x/tools v0.42.0 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20251202230838-ff82c1b0f217 // indirect
--- a/routeros/bulk_read_cache.go
+++ b/routeros/bulk_read_cache.go
@@ -35,6 +35,9 @@ type BulkReadCache struct {
 // discarded. DoBulkFetch catches this sentinel internally and retries.
 var errStaleBulkFetch = errors.New("bulk read: cache invalidated during fetch")
 
+// maxBulkFetchAttempts bounds the stale-retry loop. Each retry requires a
+// distinct intervening Invalidate to race the fetch, so 8 is well above any
+// realistic write storm and keeps a pathological case from hanging.
 const maxBulkFetchAttempts = 8
 
 func NewBulkReadCache() *BulkReadCache {
@@ -86,7 +89,6 @@ func (c *BulkReadCache) Invalidate(path string) {
 	c.mu.Unlock()
 }
 
-// currentGen returns the path's current generation under read lock.
 func (c *BulkReadCache) currentGen(path string) uint64 {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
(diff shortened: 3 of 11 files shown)


goto1134:
TBH, the performance gain is so dramatic, and the RAM price is so low, so I would consider enabling it by default. We could save users a lot of time.

PR #983, opened 2026-04-25
fix(ip_service): handle dynamic per-connection rows in RouterOS 7.20+

### LINKED ISSUE #905: Unable to import routeros_ip_service in 1.97.0 and above
**Describe the bug**
When using version 1.97.0 and above, terraform plan fails with "Error: Cannot import non-existent remote object". 

**To Reproduce**
```
resource "routeros_ip_service" "telnet" {
  numbers  = "telnet"
  port     = 23
  disabled = true
}

import {
  to = routeros_ip_service.telnet
  id = "telnet"
}
```
Running terraform plan with this config fails:

```
routeros_ip_service.telnet: Preparing import... [id=telnet]
routeros_ip_service.telnet: Refreshing state... [id=*0]

Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Cannot import non-existent remote object
│ 
│ While attempting to import an existing object to "routeros_ip_service.telnet", the provider detected that no object exists with the given id. Only pre-existing objects can be
│ imported; check that the id is correct and that it is associated with the provider's configured region or endpoint, or use "terraform apply" to create a new remote object for
│ this resource.
╵

```
**Expected behavior**
In versions up to 1.96.0 this results in a succesfull terraform plan:
```
routeros_ip_service.telnet: Preparing import... [id=telnet]
routeros_ip_service.telnet: Refreshing state... [id=telnet]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

  # routeros_ip_service.telnet will be updated in-place
  # (imported from "telnet")
  ~ resource "routeros_ip_service" "telnet" {
        address  = null
        disabled = true
        id       = "telnet"
        invalid  = true
        name     = "telnet"
      + numbers  = "telnet"
        port     = 23
        vrf      = "main"
    }

Plan: 1 to import, 0 to add, 1 to change, 0 to destroy.
```

**Additional context**
Terraform v1.14.3
RouterOS 7.11.2

SolAstrius:
## Summary

Closes #905.

RouterOS 7.20 changed `/ip service print` to surface dynamic, per-connection rows alongside each static service config. Multiple entries now share the same `name` (e.g. two `name="ssh"` rows: the static service entry plus a dynamic row for each active SSH session). The previous implementation used `name` as the canonical identifier (`MetaId: PropId(Name)`), which broke import and refresh:

- Reads with `name=ssh` matched both the static row and active connections, surfacing as `more than one resource found: name=ssh`.
- `terraform import` with a `*N` handle succeeded at the importer step but the subsequent Read still filtered by `name`, hitting the same ambiguity.
- `terraform import ... ftp` (and the other always-static services like winbox/api-ssl/www/telnet) failed with `Cannot import non-existent remote object` — surfaced as a misleading message when the multi-match handling skipped the right row.

## Approach

Switch to `.id`-based identification end-to-end and filter `dynamic=false` on every read path:

- `MetaId: PropId(Id)` — canonical identifier is the `*N` handle.
- Custom Create/Update: after `/set`, look up the static row by `name=<numbers> dynamic=false` and store its `.id`.
- Custom Read (`resourceIpServiceRead`): filter by `.id=<d.Id()> dynamic=false`.
- Custom Importer (`importIpServiceState`): accept either a `*N` handle directly, or a service name like `ssh` (resolved to `.id` via `name=<n> dynamic=false`). Backwards-compatible with the previous name-based imports — `terraform import routeros_ip_service.telnet telnet` still works.

The `dynamic=false` filter ensures per-connection rows never leak into terraform state regardless of how many sessions are active when the read fires.

## Test plan

- [x] Verified on a live RouterOS 7.21.3 router (a hAP ac²) that all 7 services (`ftp`, `ssh`, `telnet`, `winbox`, `www`, `api`, `api-ssl`) import cleanly using both bare names and `*N` handles, both with active SSH sessions causing dynamic-row duplicates and without.
- [x] Verified subsequent `tofu plan` shows no spurious drift after import.
- [x] `go vet ./routeros/...` and `go build ./...` clean.
- [ ] Existing acceptance test (`TestAccIpServiceTest_basic`) still passes — couldn't run locally without a CI RouterOS instance; relying on existing CI to confirm.

## Notes

The `numbers` field semantics are unchanged — users still write `numbers = "ssh"` to identify which service to configure. Only the internal terraform resource ID layout changed (from `"ssh"` to `"*N"`); existing state files keyed by name are migrated transparently on first refresh because the Read path now resolves names to `.id` automatically.

### DIFF
--- a/routeros/resource_ip_service.go
+++ b/routeros/resource_ip_service.go
@@ -2,6 +2,8 @@ package routeros
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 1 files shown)


rymdbullen:
Hello,
Humble question about the feasibility of a merge of this PR?

Thanks in advance

Issue #966, opened 2026-03-18
Missing field in routeros_interface_wireguard_peer (7.21)


durandguru:
**Describe the bug**
Warning: Field 'client_allowed_address' not found in the schema
│
│   with module.wireguard-s2s.routeros_interface_wireguard_peer.peer["s2s-in-event"],
│   on .terraform\modules\wireguard-s2s\wireguard-s2s\peers.tf line 1, in resource "routeros_interface_wireguard_peer" "peer":
│    1: resource "routeros_interface_wireguard_peer" "peer" {
│
│ [MikrotikResourceDataToTerraform] The field was lost during the Schema development: ▷ 'client_allowed_address': '' ◁
│
│ (and one more similar warning elsewhere)

dyegoe:
Yes, I can confirm the behaviour above with the most recent routeros 1.99.1 and RouterOS 7.21

PR #987, opened 2026-05-04
feat(routeros): add client_allowed_address field to wireguard peer

dyegoe:
- This field became configurable on RouterOS 7.21
- Introduced a new optional field `client_allowed_address` to specify CIDRs for routing on the client side.
- Added description for the field to guide users on its usage.
- Included a suppression function to handle diff presentation.

This is to address issue #966


dyegoe:
Hi, this is my first contribution. I am very open to any feedback.

I didn't update the CHANGELOG because version 1.100.0 has already been listed, but I haven't seen it released yet. I'm uncertain about the next steps.

I haven't added any tests for this new field. Additionally, I'm unsure how it behaves with older RouterOS versions that lack this field.

Maybe someone with more experience can suggest the best course of action.

PR #998, opened 2026-07-21
fix: make ip_service, ip_dns and ip_neighbor_discovery_settings importable

tkensiski:
## What

Three resources couldn't be imported cleanly. This fixes all three and adds an import test for each.

### `routeros_ip_service`
`IdType` is `Name` (the menu is fixed and name-addressed), but the resource used the generic importer, which stored the internal `.id` (e.g. `*6`). The name-keyed read then couldn't find it, and the `Required` `numbers` selector was never populated — a permanent post-import diff. A small custom importer now imports by service name (e.g. `www-ssl`), also accepts the provider-wide `name=`/`numbers=` form, and seeds `numbers` from the id so the first read is drift-free.

### `routeros_ip_dns`
The DNS settings are a singleton but had no importer, so they couldn't be imported at all. Added `ImportStatePassthroughContext` (import with `.`). The example `import.sh` and generated doc are corrected from "can not be imported" accordingly.

### `routeros_ip_neighbor_discovery_settings`
Also a singleton, but it used `ImportStateCustomContext` (which expects a `field=value` id) and failed to import. Switched to `ImportStatePassthroughContext`.

## Tests
Each resource gains an import step — `ImportStateVerify` for `ip_service`, `ImportState` for the two singletons.

---
_Disclosure: an AI assistant (Claude) helped draft this change; a human reviewed it and verified all three imports against real RouterOS hardware before submission._

### DIFF
--- a/docs/resources/ip_dns.md
+++ b/docs/resources/ip_dns.md
@@ -48,6 +48,6 @@ resource "routeros_ip_dns" "dns-server" {
 ## Import
 Import is supported using the following syntax:
 ```shell
-#The DNS Settings can not be imported. 
-#Terraform will ignore the current settings and will overwrite the current settings with the settings defined in Terraform.
+# The DNS settings are a singleton; import with '.'.
+terraform import routeros_ip_dns.test .
 ```
--- a/examples/resources/routeros_ip_dns/import.sh
+++ b/examples/resources/routeros_ip_dns/import.sh
@@ -1,2 +1,2 @@
-#The DNS Settings can not be imported. 
-#Terraform will ignore the current settings and will overwrite the current settings with the settings defined in Terraform.
+# The DNS settings are a singleton; import with '.'.
+terraform import routeros_ip_dns.test .
--- a/routeros/resource_ip_dns.go
+++ b/routeros/resource_ip_dns.go
@@ -196,6 +196,14 @@ func ResourceDns() *schema.Resource {
 			return nil
 		},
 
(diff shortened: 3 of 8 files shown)


tkensiski:
Closing for now, will reopen once I verify backwards compat with older versions

BenjaminVouillaume:
Just wanted to add a +1 on this one. I'm currently bringing my home router under Terraform and ran into the exact same two singleton import issues (`ip_dns` and `ip_neighbor_discovery_settings`) — I found this PR while searching for a solution, and it describes precisely what I was seeing.

This fix would have saved me a good amount of head-scratching, so thanks for putting it together.

Hoping it gets reopened and merged when you get the chance to finish the backwards-compat checks. 🤞

tkensiski:
@BenjaminVouillaume first off glad to know I’m not the only one! I have a few more loose ends I ran into while configuring my hardware, then need to regression test. Should have it reopened in the next few days and will have a branch for you to use while waiting on a merge here!

tkensiski:
@BenjaminVouillaume sorry this is taking a bit longer, turns out theres some bigger issues that need sorting (no updates at all for 5 months as one example as well as a discussion about how to handle min/max versions for features).

In the meantime, heres my working branch with these fixes applied and a few others as well:
https://github.com/tkensiski/terraform-provider-routeros/tree/integration

More context:
- https://github.com/orgs/terraform-routeros/discussions/958

